PR #9, opened 2024-10-29
feat: larc section rest & graphql endpoints

### LINKED ISSUE #3: Endpoint for LARC data
## Summary

Create endpoints for `larc_section` data

- [x] REST endpoint for `larc_section` data
- [x] GraphQL endpoint for `larc_section` data

andrew-wang0:
## Description

Created endpoints for `larc_section` data.

`v2/rest/larc/`: queries larc sections by filter
\+ equivalent graphql endpoint

Closes #3


alexanderl19:
We recently decided to remove the all courses endpoint in favor of a single endpoint that doesn't require parameters. Would it make sense here to mirror that?

Also @ecxyzzy ? added pagination to the courses endpoint to reduce the max response size and in memory data. The worker will error if it exceeds 128 mb. This might not be a problem with LARC since I assume there are less joins, but it may be worthwhile to consider adding pagination with a larger window.

andrew-wang0:
> We recently decided to remove the all courses endpoint in favor of a single endpoint that doesn't require parameters. Would it make sense here to mirror that?

Will do! The current `larc` endpoint already doesn't require params, so I'll just remove the `larc/all` endpoint.

ecxyzzy:
LARC should not have the same issue since it is only doing one join, and availability of historical data is much less, only going back to 2019 rather than 1990 for WebSoc. It still makes sense to remove the `/all` endpoint for the sake of consistency.

alexanderl19:
> LARC should not have the same issue since it is only doing one join, and availability of historical data is much less, only going back to 2019 rather than 1990 for WebSoc. It still makes sense to remove the `/all` endpoint for the sake of consistency.

Do you think this will be a problem in the foreseeable future? Cause adding pagination down the line would probably constitute a breaking change. Though I presume this is more akin to the WebSoc endpoint rather than the courses one now that I think about it.

ecxyzzy:
The WebSoc endpoint already requires the year and quarter to be populated, so unless the school's offering of sections per term increases by...a lot, I don't really foresee that being an issue. It might make sense to limit the LARC endpoint in the same way if we think it will also accumulate enough data to break usecases where all data is fetched.

andrew-wang0:
>  It might make sense to limit the LARC endpoint in the same way if we think it will also accumulate enough data to break usecases where all data is fetched.

Given that there is relatively minimal LARC data in the current response, it's unlikely that this would be an issue. Though, I'm assuming that most requests would want to filter by year & quarter anyways, so it might be good to implement for longevity & consistency sake.

ecxyzzy (commented):
Implementation looks good so far, good work! Once the `year` and `quarter` params are made mandatory I think we can ship :rocket:

ecxyzzy (approved):
👍

PR #12, opened 2024-10-30
feat: fuzzy SaaS

ecxyzzy:
## Description

Implement an access-controlled fuzzy search endpoint built on Postgres full text search. To that end, also implemented the logic necessary to support access-controlled endpoints, and refactored the Courses/Instructors services to allow fetching objects that are the same shape as what the API returns, but based on raw SQL filters.

The current department aliases are as follows:
- ICS -> I&C SCI
- CS -> COMPSCI
- INF -> IN4MATX
- WR -> WRITING
- ESS -> EARTHSS

## Related Issue

Closes #11.

## How Has This Been Tested?

Tested locally since we don't have a CI pipeline yet.

Here's how to get setup locally:
- Back up your local database instance by running the `dump_db.sh` script at the project root. Store the old `db.sql.gz` somewhere safe.
- Download the new database dump from [here](https://anteater-api-dump.s3.us-west-1.amazonaws.com/db-staging-12.sql.gz), put it in `packages/db`, and rename it to `db.sql.gz`.
- Recreate your local database instance by running `docker-compose down && docker-compose up --build -d`.
- Wait for seeding to complete.
- Create an API key locally by running this command. Replace `YOUR_API_KEY` with whatever you'd like to call your API key. You will need this to test locally.
`pnpm wrangler kv key put 'YOUR_API_KEY' '{"_type":"secret","resources":{"FUZZY_SEARCH":true}}' --namespace-id=4c8dd3090bea4d469cd3c66b3002a5a2 --local`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code involves a change to the database schema.
- [ ] My code requires a change to the documentation.

Issue #5, opened 2024-10-24
Websoc scraper deadlocking


andrew-wang0:

ecxyzzy:
It looks like this issue is resolved for now, but the scraper could still use some additional enhancements to make it more robust. Will fold it into my PR that fixes #4.

PR #13, opened 2024-10-31
fix: resolve issues with deadlocking

ecxyzzy:
## Description
While attempting to migrate the prod database to the latest version of the schema introduced in #12, I noticed that it failed due to a deadlock. Even after temporarily disabling the WebSoc scraper and killing all outstanding connections it would still resurface, and as it turns out limiting the connector to one connection resolved that issue.

Things are a bit murky because the old connections may not have been killed properly, but this should be a step in the right direction.

## Related Issue

Aims to address #5, but the jury is still out on whether this resolves the issue.

## How Has This Been Tested?

Tested locally, migration succeeded.

## Types of changes

## Checklist:

### DIFF
--- a/packages/db/drizzle.config.ts
+++ b/packages/db/drizzle.config.ts
@@ -2,13 +2,8 @@ import "dotenv/config";
 
 import { defineConfig } from "drizzle-kit";
 
-const url = process.env.DB_URL;
-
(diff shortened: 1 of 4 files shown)


andrew-wang0 (approved):
Works locally! Let's see how it does on prod 👌

PR #21, opened 2024-11-04
feat(search): implement resultType flag

ecxyzzy:
## Description

Add the `resultType` flag that dictates what types of results are returned by the search endpoint. Can be `course` or `instructor`. Behavior is the same as the endpoint currently if not provided.

## How Has This Been Tested?

Tested locally with queries using the new flag.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/search.ts
+++ b/apps/api/src/schema/search.ts
@@ -9,6 +9,7 @@ export const searchQuerySchema = z.object({
     .default(100)
     .refine((x) => x <= 100, "Page size must be smaller than 100"),
   skip: z.coerce.number().default(0),
+  resultType: z.union([z.literal("course"), z.literal("instructor")]).optional(),
 });
 
 export const searchResultSchema = z.discriminatedUnion("type", [
--- a/apps/api/src/services/search.ts
+++ b/apps/api/src/services/search.ts
@@ -110,7 +110,7 @@ export class SearchService {
         id: instructor.ucinetid,
         rank: sql`TS_RANK(${INSTRUCTORS_WEIGHTS}, ${query})`.mapWith(Number),
       })
-      .from(course)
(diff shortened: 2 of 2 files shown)


andrew-wang0 (commented):
Looks good! 👌

There is an oddity though that I didn't catch the first time. The query builder here

```ts
// /apps/api/src/services/search.ts
const toQuery = (query: string) =>
  sql`TO_TSQUERY('english', WEBSEARCH_TO_TSQUERY('english', ${splitAtLastNumber(query)})::TEXT || ':*')`;
```
looks to do some shenanigans to escape stop words, conjunctions, and other symbols

This causes queries such as the following to fail:
`query=i`
`query=or` (maybe beginning a search for organic chem)
`query=the` (maybe beginning a search for theater)

with the following error
```json
{
  "ok": false,
  "message": "syntax error in tsquery: ':*'"
}
```

And queries such as the following do not return any relevant courses:
`query=non-government` (SOC SCI 152A is titled `Non-Government Organization (NGO) Fundamentals`, but the dash symbol `-` negates the word government)

ecxyzzy:
Good catch, thanks! I think this is by design for the `TO_TSQUERY` function, but is definitely not desirable for our uses. Will look into this and put up an additional commit.

ecxyzzy:
Should be fixed now!

For posterity, the `non-government` query was probably exploding because we passed the node into another `TO_TSQUERY` call which expanded it unnecessarily; when I was testing in `psql`, the negation would not happen with `WEBSEARCH_TO_TSQUERY` unless the hyphen begins a lexeme. Escaping the hyphen looks like it helped in this case, and queries containing only stop words will now just return an empty array rather than erroring.

andrew-wang0 (approved):
Looks Good! :ok_hand:

PR #23, opened 2024-11-09
perf: optimize courses and instructors routes

ecxyzzy:
## Description

Optimize performance of courses/instructors by using materialized views, where each view is built using a query that performs JSON aggregation on the DBMS.

Also modify the WebSoc scraper to update the materialized view every time it completes a scrape.

## How Has This Been Tested?

Benchmarked locally with Postman, see screenshot.

The numbers are an apples-to-apples comparison when tested locally so in production it will definitely be higher, but should still be a pretty significant improvement since we've also cut down on the number of round trips.

image

## Types of changes

## Checklist:


andrew-wang0 (approved):
Blazingly fast  :+1: :sunglasses: :fire:

PR #24, opened 2024-11-09
feat: batch endpoints for courses/instructors

ecxyzzy:
## Description

Add support for fetching courses/instructors in a bulk, given a set of IDs.

## Motivation and Context

We have been able to support batch queries for a long time but have never done so (mostly due to my poor planning). This should remove the need for consumers to query our API _N_ times for _N_ things, if they already know exactly what they want.

## How Has This Been Tested?

Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/courses.ts
+++ b/apps/api/src/schema/courses.ts
@@ -41,6 +41,10 @@ export const coursesPathSchema = z.object({
     .openapi({ param: { name: "id", in: "path" } }),
 });
 
+export const batchCoursesQuerySchema = z.object({
+  ids: z.string().transform((xs) => xs.split(",")),
+});
+
 export const coursesQuerySchema = z.object({
   department: z.string().optional(),
   courseNumber: z.string().optional(),
--- a/apps/api/src/schema/courses.ts
+++ b/apps/api/src/schema/courses.ts
@@ -42,7 +42,10 @@ export const coursesPathSchema = z.object({
 });
 
 export const batchCoursesQuerySchema = z.object({
-  ids: z.string().transform((xs) => xs.split(",")),
+  ids: z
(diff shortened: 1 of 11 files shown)


apps/api/src/schema/courses.ts
```diff
@@ -41,6 +41,10 @@ export const coursesPathSchema = z.object({
     .openapi({ param: { name: "id", in: "path" } }),
 });
 
+export const batchCoursesQuerySchema = z.object({
+  ids: z.string().transform((xs) => xs.split(",")),
```
andrew-wang0, line 45:
```suggestion
  ids: z
    .string({ message: "Parameter 'ids' is required" })
    .transform((xs) => xs.split(","))
    .openapi({ example: "COMPSCI161,COMPSCI162" }),
```

apps/api/src/schema/instructors.ts
```diff
@@ -6,6 +6,10 @@ export const instructorsPathSchema = z.object({
     .openapi({ param: { name: "ucinetid", in: "path" } }),
 });
 
+export const batchInstructorsQuerySchema = z.object({
+  ucinetids: z.string().transform((xs) => xs.split(",")),
```
andrew-wang0, line 10:
```suggestion
  ucinetids: z
    .string({ message: "Parameter 'ucinetids' is required" })
    .transform((xs) => xs.split(","))
    .openapi({ example: "mikes,klefstad" }),
```

andrew-wang0 (approved):
:ok_hand:

apps/api/src/schema/courses.ts
ecxyzzy: Good call, thanks for catching this!

andrew-wang0 (approved):
:ok_hand:

Issue #2, opened 2024-10-24
Endpoint for study room availability


andrew-wang0:
## Summary

Create endpoints for `study_room` data

- [ ] REST endpoint for `study_room` data
- [ ] GraphQL endpoint for `study_room` data

alexanderl19:
Wooo nice! Also are we not yet returning availability? Tried a couple ids and didn't get the slots field.

ecxyzzy:
> Wooo nice! Also are we not yet returning availability? Tried a couple ids and didn't get the slots field.

Yeah we're not returning availability yet - tracking the follow-up in #46

PR #29, opened 2024-11-12
feat: add rest and graphql endpoints for study rooms

sanskarm7:
## Description

Created GraphQL and REST endpoints to provide study room data, namely:

- allStudyRooms: Returns the id, name, capacity, location, and techEnhanced fields for each study room in the database.
- studyRoomById: Returns the id, name, capacity, location, techEnhanced, description, and directions for a specific room, if available.
- studyRoomsByFilters: Filters and returns study rooms based on query parameters (location, capacityMin, capacityMax, isTechEnhanced), providing the same fields as allStudyRooms.

Added description and directions fields to the studyRoomById endpoint to provide more details for each specific room.

## Related Issue

resolves issue #2 

## Motivation and Context

These endpoints provide structured access to study room data in both REST and GraphQL formats, fulfilling the requirements outlined in issue #2. These changes allow users to access study room details as a whole, retrieve data for a specific room, or filter rooms based on criteria like location and capacity. This functionality is essential for making study room data available through the API in a scalable and flexible way.

## How Has This Been Tested?

Tested locally on a development server using Postman. Verified:

REST Endpoints:

Access to /v2/rest/study-rooms/all returns all rooms.
Access to /v2/rest/study-rooms/{id} returns data for a specific room by ID.
Access to /v2/rest/study-rooms with query parameters returns filtered results.

GraphQL Queries:

Querying allStudyRooms returns a list of all study rooms with the expected fields.
Querying studyRoom(id: "<room_id>") returns details for a specific room, including description and directions if they are present.
Querying studyRooms(query: { location: "Science Library", capacityMin: 4 }) returns filtered results based on location and capacity.

## Types of changes

## Checklist:

- [X] My code requires a change to the documentation.


apps/api/src/rest/index.ts
```diff
@@ -23,6 +24,7 @@ restRouter.route("/ping", pingRouter);
 restRouter.route("/search", searchRouter);
 restRouter.route("/websoc", websocRouter);
 restRouter.route("/week", weekRouter);
+restRouter.route("/study-rooms", studyRoomsRouter);
```
ecxyzzy, line 27: REST resources should be camelCase
```suggestion
restRouter.route("/studyRooms", studyRoomsRouter);
```

apps/api/src/rest/routes/study-rooms.ts
```diff
@@ -0,0 +1,129 @@
...
+      content: { "application/json": { schema: errorSchema } },
+      description: "Server error occurred",
+    },
+  },
+});
+
+studyRoomsRouter.get(
+  "*",
+  productionCache({ cacheName: "study-room-api", cacheControl: "max-age=86400" }),
+);
```
ecxyzzy, line 95: We should probably not be caching this endpoint since it'll be providing real-ish time data of study room availability.

apps/api/src/schema/study-rooms.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const studyRoomSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  capacity: z.number().int(),
+  location: z.string(),
+  description: z.string().optional(),
+  directions: z.string().optional(),
+  techEnhanced: z.boolean(),
+  slots: z.array(slotSchema).optional(),
```
ecxyzzy, line 18: Is there a reason the slots array is nullable here but not in the GraphQL schema?

apps/api/src/services/study-rooms.ts
```diff
@@ -0,0 +1,41 @@
+import type { database } from "@packages/db";
+import { and, eq, gte, lte } from "@packages/db/drizzle";
+import { studyRoom } from "@packages/db/schema";
+
+type StudyRoomsServiceInput = {
+  location?: string;
+  capacityMin?: number;
+  capacityMax?: number;
+  isTechEnhanced?: boolean;
+};
```
ecxyzzy, line 10: Can be replaced with `z.infer<typeof studyRoomsQuerySchema>`

apps/api/src/services/study-rooms.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class StudyRoomsService {
+  constructor(private readonly db: ReturnType<typeof database>) {}
+
+  async getAllStudyRooms() {
+    return this.db
+      .select()
+      .from(studyRoom)
+      .then((rows) => rows);
```
ecxyzzy, line 19: This seems unnecessary?

apps/api/src/services/study-rooms.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (input.capacityMin) conditions.push(gte(studyRoom.capacity, input.capacityMin));
+    if (input.capacityMax) conditions.push(lte(studyRoom.capacity, input.capacityMax));
+    if (input.isTechEnhanced !== undefined)
+      conditions.push(eq(studyRoom.techEnhanced, input.isTechEnhanced));
+
+    return this.db
+      .select()
+      .from(studyRoom)
+      .where(and(...conditions))
+      .then((rows) => rows);
```
ecxyzzy, line 39: Same here

ecxyzzy (changes requested):
We're moving away from implementing `/all` routes for resources in favor of paginated root routes that allow empty filters. Feel free to reference the courses/instructors routes for how to implement such an endpoint, or ask me or Andrew for help if necessary.

Other than that looks decent overall, left comments for minor nits though. Great work 💪

ecxyzzy:
Oh also as a side note please make sure to make the PR name compliant with conventional commits before merging. Thanks :+1:

apps/api/src/services/study-rooms.ts
sanskarm7: I'm not understanding how this part is unnecessary. What is favored instead of it? I get that since we shouldn't be supporting an "all" endpoint, that the earlier segment should be removed, but for this one, the conditions map what should be returned. I believe "then" can be removed if async is used, to create something like this:
   return await this.db
      .select()
      .from(studyRoom)
      .where(and(...conditions))
      
 What is the best solution?

ecxyzzy: Sorry I should have been more clear - I meant that the `.then((rows) => rows)` part seems unnecessary because no data processing is happening, so it could just be left out.

ecxyzzy (approved):
LGTM, great work on this!

We should look into returning the available slots as a potential followup though, as this PR doesn't seem to do that.

PR #30, opened 2024-11-12
feat: properly type prerequisite tree

### LINKED ISSUE #28: Investigate possibility of properly typing prerequisite tree
Currently the attributes in the prerequisite tree object's OpenAPI spec are typed as `any[]`. This is because hono-zod-openapi doesn't support `ZodLazy`, which is necessary for self-referential datatypes like the prerequisite tree.

ecxyzzy:
## Description

Revert `prerequisiteTreeSchema` to what it was originally. Also, override the OpenAPI metadata on it so it doesn't cause any errors, and set the metadata properly so consumers can receive the correct type data.

Honestly this was a lot less involved than I expected.

## Related Issue

Closes #28.

## Motivation and Context

Now we won't have to hardcode the prereq tree type in the API's consumers.

## How Has This Been Tested?
- Run dev server locally.
- Use [`openapi-typescript`](https://github.com/openapi-ts/openapi-typescript) to generate a types file from the dev server.
- Paste the following snippet into another TS file and play around with it, ensuring that the `PrerequisiteTree` type behaves as expected.
```ts
// assuming the output is named "anteater-api-types.ts"
import { components } from "./anteater-api-types";

type PrerequisiteTree = components["schemas"]["prereqTree"];
```

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/rest/routes/courses.ts
+++ b/apps/api/src/rest/routes/courses.ts
@@ -6,6 +6,8 @@ import {
   coursesPathSchema,
   coursesQuerySchema,
   errorSchema,
+  prerequisiteSchema,
+  prerequisiteTreeSchema,
   responseSchema,
 } from "$schema";
 import { CoursesService } from "$services";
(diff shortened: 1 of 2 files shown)


andrew-wang0 (approved):
:ok_hand:

PR #33, opened 2024-11-16
fix(key-manager): misc

ecxyzzy:
## Description

- Pin dependencies.
- Resolve issue with build scripts by moving entrypoint into dev/deploy scripts.
  - Previously there was an error here caused by a circular dependency. The postinstall script needed the worker-next artifact because it's defined in the `wrangler.toml`. However, it hasn't been built yet, and can't be built because the dependencies haven't been installed yet.
- Extract key types into its own package and have both API and key manager depend on it.
- Make list of resources in key types `as const`.
- Fixup biome/lefthook config so tsx files are included and Next artifacts are excluded.
- Run biome on key-manager code and fix lint/format errors.
- Remove `isAdmin` helper as having special attributes does not imply that a key "is" an "admin key". Rather, it just means that someone in leadership has granted them an exception.

## How Has This Been Tested?

Tested locally

## Types of changes

## Checklist:


andrew-wang0 (approved):
Epic fixes :pray:

PR #44, opened 2024-11-24
chore: upgrade to node 22

### LINKED ISSUE #32: Upgrade to Node 22 and clean up relevant code
Node.js 22 has just hit LTS, bringing with it some additional goodies like support for [`Iterator.toArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/toArray) and [`Iterator.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map).

We're also no longer constrained to Node 20, since Cloudflare has confirmed that using their codegen for types (which we are) means we do not need to pin the `@types/node` package to v20, so we can actually upgrade to Node 22.

## Context

This will be useful in cleaning up components of our data pipelines, since we use the following pattern a lot:
```typescript
const theMap = new Map<string, unknown>();

const res = Array.from(theMap.keys()).map((key) => { ... });
```

Which can then be converted into this pattern:
```typescript
const theMap = new Map<string, unknown>();

// toArray can even be omitted in cases where the function consuming res accepts an Iterator
const res = theMap.keys().map((key) => { ... }).toArray();
```

## Action Items
- [ ] Set version in `.nvmrc` to 22.11.0
- [ ] Install latest `@types/node` at workspace level
- [ ] Remove `@types/node` at package level
- [ ] Refactor code using `Array.from`
- [ ] Verify that nothing broke

andrew-wang0:
## Description

Upgrade to node 22 and refactor usages of `Array.from` to `iterator.toArray()`

## Related Issue

closes #32 

## How Has This Been Tested?

Ran scrapers locally and tested changed services by hitting endpoints

## Types of changes

## Checklist:


apps/api/src/services/search.ts
```diff
@@ -126,12 +130,14 @@ export class SearchService {
...
-      results: Array.from(results.entries())
+      results: results
+        .entries()
         .map(([key, rank]) => ({
           type: "instructor" as const,
           result: getFromMapOrThrow(instructors, key),
           rank,
         }))
+        .toArray()
         .slice(input.skip, input.skip + input.take),
```
ecxyzzy, line 141: Instead of doing this, maybe consider using [`Iterator.drop`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/drop) and [`Iterator.take`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/take), since that way we can avoid having the iterator return elements we don't want.

apps/data-pipeline/degreeworks-scraper/package.json
```diff
@@ -12,7 +12,5 @@
     "cross-fetch": "4.0.0",
     "jwt-decode": "4.0.0"
   },
-  "devDependencies": {
-    "@types/node": "20.8.3"
-  }
+  "devDependencies": {}
```
ecxyzzy, line 15: This can be safely removed.

apps/data-pipeline/grades-importer/package.json
```diff
@@ -13,7 +13,5 @@
     "@packages/stdlib": "workspace:*",
     "node-xlsx": "0.24.0"
   },
-  "devDependencies": {
-    "@types/node": "20.8.3"
-  }
+  "devDependencies": {}
```
ecxyzzy, line 16: Can be removed

packages/stdlib/package.json
```diff
@@ -8,7 +8,5 @@
   "scripts": {
     "check:types": "tsc -p ./tsconfig.json -noEmit"
   },
-  "devDependencies": {
-    "@types/node": "20.8.3"
-  }
+  "devDependencies": {}
```
ecxyzzy, line 11: Can be removed

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -536,12 +536,16 @@ const doDepartmentUpsert = async (
...
-          .filter(notNull),
+          .filter(notNull)
+          .toArray(),
       ),
     );
-    const instructorAssociationsToInsert = Array.from(sectionsToInstructors.entries())
+    const instructorAssociationsToInsert = sectionsToInstructors
+      .entries()
+      .toArray()
       .flatMap(([k, names]) =>
```
ecxyzzy, line 549: Could probably use `Iterator.flatMap` then `toArray`

apps/api/src/services/search.ts
andrew-wang0: Epic, did not know that was a thing :star_struck:

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -637,7 +641,9 @@ const doDepartmentUpsert = async (
     await tx
       .insert(websocSectionMeetingToLocation)
       .values(
-        Array.from(meetingsToLocations.entries())
+        meetingsToLocations
+          .entries()
+          .toArray()
           .map(([meetingId, v]) => {
```
ecxyzzy, line 647: Sorry I'm dumb and missed this :facepalm: This can be `map` then `toArray` as well

apps/api/src/services/search.ts
```diff
@@ -96,12 +98,14 @@ export class SearchService {
...
-      results: Array.from(results.entries())
+      results: results
+        .entries()
         .map(([key, rank]) => ({
           type: "course" as const,
           result: getFromMapOrThrow(courses, key),
           rank,
         }))
+        .toArray()
         .slice(input.skip, input.skip + input.take),
```
ecxyzzy, line 109: Also missed this - can be converted to .drop and .take

ecxyzzy (approved):
:rocket:

PR #57, opened 2024-12-16
fix(search): count is a nonnegative int

### LINKED ISSUE #56: Count of search results is mistyped
## Expected Behavior

The number of results from a search result is a nonnegative integer.
This makes mathematical sense.

## Current Behavior

This is typed generically as a number.
In JS, this is a 64-bit float and is therefore overly ambiguous.

## Possible Solution

Retype this as an integer.

## Steps to Reproduce (for bugs)

N/A

## Context

i am very picky about types

## Your Environment

N/A

laggycomputer:
## Description

Retype the type of the count property on the search for instructor/course endpoint as an integer.

## Related Issue

Resolves #56.

## Motivation and Context

As explained in the issue, this is the correct type for this value from a mathematical standpoint.
This is technically a breaking change but:
* It only affects users where integers and floats are not the same (i.e. not JS/TS).
* It only affects people who somehow depended on incorrect behavior.

## How Has This Been Tested?

Local testing was done to ensure neither GraphQL nor REST are broken by this change.
Note that token checks were disabled in this local deployment to simplify local testing.

Observe that Postman introspects the new type correctly and the endpoint works as expected:
image

REST continues to work:
Screenshot_20241216_154750

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/graphql/schema/search.ts
+++ b/apps/api/src/graphql/schema/search.ts
@@ -7,7 +7,7 @@ type SearchResult @cacheControl(maxAge: 86400) {
 }
 
 type SearchResponse {
-    count: Float!
+    count: Int!
     results: [SearchResult!]!
 }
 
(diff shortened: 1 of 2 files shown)


andrew-wang0 (approved):
Looks good 🙏

Issue #31, opened 2024-11-12
Resolve empty enrollment history entries


ecxyzzy:
## Expected Behavior
The enrollment history endpoint should not return entries for terms for which we do not have enrollment data.

## Current Behavior
The enrollment history endpoint returns entries for terms for which we do not have enrollment data. In these entries, the time series data are all empty arrays.

This can lead to unintentional behavior in consumers if they assume that any entry will have at least one datapoint.

PR #64, opened 2025-01-06
fix: resolve empty enrollment history entries

waterkimchi:
## Description
Fixed the issue where the enrollment history endpoint was returning entries for terms without enrollment data. Now, the endpoint only returns entries for terms with actual enrollment data, ensuring that each entry has at least one data point.

## Related Issue
This pull request addresses issue [#31](https://github.com/icssc/anteater-api/issues/31).

## Motivation and Context

This change is required to prevent unintentional behavior in consumers of the enrollment history endpoint. By ensuring that only terms with enrollment data are returned, we avoid empty data arrays that could cause issues or unexpected behavior in downstream applications.
## How Has This Been Tested?

Tested on a local deployment by using the pnpm environment. Minor changes. None affects other code.
Before(includes the empty data sections):
Screenshot 2025-01-06 at 09 54 58
After(removed the empty data sections):
Screenshot 2025-01-06 at 09 55 21

## Types of changes

## Checklist:


apps/api/src/services/enrollment-history.ts
```diff
@@ -159,7 +159,20 @@ export class EnrollmentHistoryService {
         section.statusHistory.push(row.status ?? "");
       }
     }
-    return transformedSectionRows
+    const filteredSections = Array.from(transformedSectionRows.values()).filter((section) => {
```
andrew-wang0, line 162: The newest versions of node added some methods from `Array` onto `Iterator`, so we can preform `filter` directly on `transformedSectionRows.values()`

```suggestion
    const filteredSections = transformedSectionRows.values().filter((section) => {
```

apps/api/src/services/enrollment-history.ts
```diff
@@ -159,7 +159,20 @@ export class EnrollmentHistoryService {
...
+        section.waitlistHistory.length > 0 ||
+        section.waitlistCapHistory.length > 0 ||
+        section.requestedHistory.length > 0 ||
+        section.newOnlyReservedHistory.length > 0 ||
+        section.statusHistory.length > 0
+      );
+    });
+
+    return filteredSections
       .values()
```
andrew-wang0, line 176: Since thre previous change keeps `filteredSections` as an `Iterator` we can omit the `.values()` call here.

```suggestion
```

andrew-wang0 (changes requested):
Great work! Just a few syntactic changes and it should be good to go :saluting_face:

apps/api/src/services/enrollment-history.ts
```diff
@@ -159,7 +159,20 @@ export class EnrollmentHistoryService {
...
+    const filteredSections = Array.from(transformedSectionRows.values()).filter((section) => {
+      return (
+        section.finalExam.length > 0 ||
+        section.maxCapacityHistory.length > 0 ||
+        section.totalEnrolledHistory.length > 0 ||
+        section.waitlistHistory.length > 0 ||
+        section.waitlistCapHistory.length > 0 ||
+        section.requestedHistory.length > 0 ||
+        section.newOnlyReservedHistory.length > 0 ||
+        section.statusHistory.length > 0
```
andrew-wang0, line 171: Good work here! But there is a slight optimization to be made. Although boolean comparisons are relatively fast, from the above lines:

```ts
section.dates.push(row.createdAt.toISOString().split("T")[0]);
section.maxCapacityHistory.push(row.maxCapacity.toString(10));
section.totalEnrolledHistory.push(row.numCurrentlyTotalEnrolled?.toString(10) ?? "");
section.waitlistHistory.push(row.numOnWaitlist?.toString(10) ?? "");
section.waitlistCapHistory.push(row.numWaitlistCap?.toString(10) ?? "");
section.requestedHistory.push(row.numRequested?.toString(10) ?? "");
section.newOnlyReservedHistory.push(row.numNewOnlyReserved?.toString(10) ?? "");
section.statusHistory.push(row.status ?? "");
```

we know that all of these lists on `section` should be the same length, since we always push together into each list. 

The current implementation is well optimized for entries with valid time-series data, since checking  only `section.finalExam.length` short circuits the expression to be true.  However, the chaining of `||` operators means that entries without any time-series data (the ones we are trying to remove) will run through all 8 comparisons.

Because each list is the same length, you check just one list and get the same result. For example, checking `section.dates.length` could help make it clear that you're filtering by sections with valid time-series data while also having a small performance improvement.

andrew-wang0 (approved):
Looks Good!

Issue #74, opened 2025-01-10
Consolidate date query schema across endpoints


andrew-wang0:
Follow up to #73 

## Possible Solution

See #73, but apply this to the `date` param

laggycomputer:
The choices currently available are `z.coerce.date()` (WebSoc endpoints) and `z.string().openapi({ format: "date-time" })` (study room endpoints). Note that all dates are simply marked as strings in GraphQL.

Issue #46, opened 2024-11-24
Return availability in studyRoom endpoint


ecxyzzy:
## Expected Behavior

`studyRoom` endpoints should return the room's availability.

Should consider limiting the availability returned if not querying one room, otherwise may result in too much data being returned at once.

## Current Behavior
`studyRoom` endpoints do not return the room's availability.

## Context
Follow-up issue to #29

PR #75, opened 2025-01-12
feat(study-rooms): serve slots data

laggycomputer:
## Description

Adds study room slots and availability to study room endpoint.

As previously implemented in the data pipeline, availability is available in on the days between the day of the current time in UCI's timezone (`America/Los_Angeles`) and exactly 3 days later, inclusive. Data is updated every 2 minutes.

This PR simply exposes that data. We use ISO 8601 dates ~~(most precisely, `z.string().datetime({ local: true })`) which are given with the `Z` indicating UTC but should be interpreted as being in UCI's timezone~~. We ask Postgres to interpret the `TIMESTAMP NO TIMEZONE` in `America/Los_Angeles` and format as zoned ISO 8601, avoiding adding a JS timezone library.

**Note**: The `slots` field on returned study rooms is now guaranteed to exist. This is not a breaking change.

We use a materialized view to precompute the JSON structure of the response. The refreshing of this view is relatively light and is done as part of the study room and slot scraping process every 2 minutes. However, it is not completely trivial; see below demonstrating that it produces a significant speed improvement per request which will accrue over time. **This design decision is a database schema change.** Since we have this JSON structure beforehand, we need only `SELECT` with conditions when serving a response.

## Related Issue

Feature update as requested in #46. ~~Also closes the inconsistency raised in #74 since study room data was one of the offenders and is within the jurisdiction of this PR.~~ It doesn't.

## Motivation and Context

>be me
>want study room slots
>no study room slots

mfw

## How Has This Been Tested?

Tested with Postman on a local deployment.

### Old screenshots (before timezone correction):

REST is functional:
Screenshot_20250111_192159

GraphQL is functional:
Screenshot_20250111_192242

### New screenshots (latest):

Screenshot_20250114_145739

Screenshot_20250114_150118

## Types of changes

## Checklist:


andrew-wang0 (changes requested):
As discussed, the study room slots `start` and `end` times being displayed like UTC time while being in PST/PDT could be confusing. 

We've resolved to use true UTC time instead :ok_hand:

laggycomputer:
To be precise, I'm going to specify the correct UTC offset in the response.

laggycomputer:
Without fully redoing tests, the correct offset is now specified:
Screenshot_20250113_211320

We add `moment-timezone` with this PR; this is a change in the convention of this project to use vanilla `Date` and may require some further work to re-standardize.

laggycomputer:
Demoting to draft; I've just remembered Postgres supports JSON aggregation and benchmarking should be done to determine whether it is a better (faster) approach.

andrew-wang0:
Wow mindblowing work on this, will dig in to test

apps/api/src/services/study-rooms.ts
andrew-wang0, line 1: There's a few lint/format errors on this file, not sure why commit hooks didn't catch them :thinking: 

`pnpm check:biome` should fix it

andrew-wang0 (changes requested):
Looks good! A few format errors but other than that it's good to go

packages/db/migrations/0008_study-room-slots-materialized-view.sql
andrew-wang0, line 1: Another quick nitpick here, our migrations so far have followed snake case naming. So a quick rename of this sql file and changing the entry in the migration journal should work

andrew-wang0 (changes requested):
Another quick nitpick

packages/db/migrations/0008_study-room-slots-materialized-view.sql
laggycomputer: Will redo the migration by tomorrow morning.

andrew-wang0 (approved):
Looks good!

PR #85, opened 2025-01-15
fix(study-rooms): specify and serve UCI timezone for slots

### LINKED ISSUE #84: [Study Room] Study room slot openapi `start` and `end`
## Current Behavior

The current `start` and `end` openapi examples on the study room endpoint are outdated

laggycomputer:
## Description

Followup to #75, fixing study room slot times being rendered in Postgres's local timezone.
We simply set the timezone on the connection acquired by the scraper to `America/Los_Angeles`.

## Related Issue

Resolves #84. We both specify the intended behavior from #75 in the example and actually fulfill it.

## How Has This Been Tested?

Tested by setting Postgres's timezone to `America/New_York`, running the scraper, then testing the API request.

Despite forcing Postgres's global timezone to change, the timezone is correct since we set it on the session which takes precedence over the global setting:

Screenshot_20250115_110206

We know this is effective because setting a different timezone for the scraper's SQL session uses that timezone instead:

Screenshot_20250115_110404

So if we force `America/Los_Angeles` on the scraper session, it will give the correct offset regardless of Postgres or system timezone.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/study-rooms.ts
+++ b/apps/api/src/schema/study-rooms.ts
@@ -2,8 +2,8 @@ import { z } from "@hono/zod-openapi";
 
 export const slotSchema = z.object({
   studyRoomId: z.string(),
-  start: z.string().datetime({ offset: true }),
-  end: z.string().datetime({ offset: true }),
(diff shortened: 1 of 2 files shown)


andrew-wang0 (approved):
LGTM :saluting_face:

Issue #80, opened 2025-01-13
Add descriptive response message when accessing base route "/"


andrew-wang0:
## Expected Behavior

Since it is likely for users to hit the base route (eg when running `dev`), we could add a friendlier response for specifically the base route `/`

## Current Behavior

Simple 404 response `{"ok":false,"message":"The requested resource could not be found."}` that is not informative

## Possible Solution

This could instruct them to use `/reference` or `/v2/rest/...` or `/v2/graphql`

laggycomputer:
Do you think this would require an OpenAPI spec?

andrew-wang0:
@laggycomputer No, I believe sanctioning the route in `i/apps/api/src/index.ts` should suffice

rsahyoun:
@andrew-wang0 May I take on this task if that is alright with you.

andrew-wang0:
@rsahyoun you got it!

andrew-wang0:
This might be better implemented as a check if the user agent is a browser, and redirecting to `/reference`, skipping the step of having to tell the user. Since, if the user was accessing the base route, they most likely wanted to go to the reference.

rsahyoun:
Ok good to know, I'll keep that in mind when I approach this issue.

PR #88, opened 2025-01-19
feat(base-route): Redirecting user on browser to reference path

rsahyoun:
## Description

This PR implements a base route handler middleware that automatically redirects browser-based requests from the root path ("/") to "/reference". The middleware detects common browser user agents and only performs the redirect for browser-based requests, allowing API requests to proceed normally. A list of common browsers was used in the middleware to help distinguish between a user in a browser and a user who is not. 

## Related Issue

Link to issue #80 

## Motivation and Context

This change improves the user experience by automatically redirecting browser users to the API reference documentation when they visit the root URL. This ensures that human users find the documentation easily while maintaining the root endpoint's availability.

## How Has This Been Tested?

Testing performed:
- Verified browser detection for popular browsers with Chrome and Microsoft Edge
- Confirmed redirect behavior when accessing root path via browser
- Verified API requests (non-browser) still work as expected (Using Postman)
- Tested that other routes are unaffected by the middleware

Screenshot 2025-01-19 001624

Screenshot 2025-01-19 001637

Screenshot 2025-01-19 001655

Screenshot 2025-01-19 001708

## Types of changes

## Checklist:


apps/api/src/index.ts
```diff
@@ -50,6 +51,9 @@ app.notFound((c) =>
 
 // Middleware configuration
 
+//Check before all other middleware
```
laggycomputer, line 54: Space after `//` by convention. Also not sure the comment is necessary but that's opinionated.

apps/api/src/index.ts
```diff
@@ -50,6 +51,9 @@ app.notFound((c) =>
...
+app.use("/", baseRouteHandler);
```
laggycomputer, line 55: Name is a little vague; if you can find one that succinctly and accurately conveys the purpose of this middleware (e.g. `baseRouteRedirect`, `redirectBrowsers`, `redirectBrowsersToDocs`) then use that.

apps/api/src/middleware/base-route-handler.ts
```diff
@@ -0,0 +1,27 @@
+import { createMiddleware } from "hono/factory";
+
+export const baseRouteHandler = createMiddleware(async (c, next) => {
+  if (c.req.path === "/") {
```
laggycomputer, line 4: This check is redundant to the path specified used to mount the middleware. Use one or the other (preferably the mount path).

apps/api/src/middleware/base-route-handler.ts
```diff
@@ -0,0 +1,27 @@
...
+    const userAgent = c.req.header("user-agent") || "";
+
+    // Common browsers
+    const browserIdentifiers = [
```
laggycomputer, line 8: Extract this array into the module scope and cast it `as const`. This is a compiler hint that this array can be constructed and stored once globally and is deeply immutable.

laggycomputer (changes requested):
Excellent decision to use middleware. Some minor changes, some of which are more subjective.

apps/api/src/index.ts
rsahyoun: ok I just got rid of the comment to maintain the order of the syntax within the file.

apps/api/src/index.ts
rsahyoun: I chose to change the file name and the middleware to redirectBrowsersToDocs in order to be more specific on the changes.

apps/api/src/middleware/base-route-handler.ts
rsahyoun: Got rid of the redundant if check and use the mount path as the only checker

apps/api/src/middleware/base-route-handler.ts
rsahyoun: Extracted the array and put it into module scope and cast it as const, everything still works as well as before.

laggycomputer (approved):

andrew-wang0 (approved):
Good work!

PR #89, opened 2025-01-20
chore: create library for year query format

### LINKED ISSUE #73: Consolidate year query schema across endpoints
## Expected Behavior

All year query schemas  specifying a `year` parameter should borrow from the same definition, such as 

```ts
  year: z
    .string({ message: "Parameter 'year' is required " })
    .regex(/^\d{4}$/, { message: "Invalid year provided" })
    .openapi({ example: "2024" }),
 ```

## Current Behavior

Currently, there is a separate definition for each schema, with varying implementations despite achieving the same result.

Eg:
`websocQuerySchema` defines it as 
```ts
  year: z
    .string({ message: "Parameter 'year' is required " })
    .length(4, { message: "Parameter 'year' must have length 4" })
```

but `calendarQuerySchema` defines it as
```ts
year: z
    .string({ message: "Parameter 'year' is required" })
    .length(4, { message: "Parameter 'year' must have length 4" })
    .refine((x) => !Number.isNaN(Number.parseInt(x, 10)), {
      message: "Parameter 'year' must be an integer",
    })
    .openapi({ param: { name: "year", in: "query" }, example: "2024" }),
 ```

## Possible Solution

For the schemas in `/apps/api/src/schema` that specify a year parameter, consolidate them into a single definition in a new file, such as  `/apps/api/src/schema/lib/year.ts`.

Also, consider cases where `string` to `number` coercions for `year` are still necessary in the schema, such as in `weekQuerySchema`.

### PR DESCRIPTION
**Pull request by sagardhunna**
## Description

- Year schema library verifies that the value of year is an integer of length 4 and returns error statements if not
- In certain use cases it is optional so it remains optional in only those situations, along with situations where it must be converted to a number.

## Related Issue

Closes #73 

## Motivation and Context

- Simplifies code making it easier to read.

## How Has This Been Tested?

- Ran locally with comparisons using postman.

## Types of changes

## Checklist:

### CODE REVIEW

**Comment by andrew-wang0** [comment]
Looks good! One small nitpick, could you rename the PR title to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)

Something like `chore: create library for year query format` 

@laggycomputer will dig in and comment more on your code

File: apps/api/src/schema/lib/year.ts [hunk 1]
```diff
@@ -0,0 +1,22 @@
+import { z } from "@hono/zod-openapi";
+
+/**
+ * Confirms string for year is in valid format
+ * Must be 4 characters
+ * Could be String/Integer upon confirmation
+ *
+ * 1. Checks if it is a string
+ * 2. Checks length of string is 4
+ * 3. Checks if it is an integer
```
laggycomputer, line 10: Same as above; this is not actually what the parsing does! A less careful reader might not realize the distinction and trust your comment, relying on an incorrect understanding and writing unsound code based on it.

laggycomputer (changes requested):
Few little things, including minor oversights. It's important to get this correct because it's a very important little snippet.

apps/api/src/schema/lib/year.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Leaving a default openapi schema, if other schemas are adding to it, leave that separate.
+ *
+ */
+
+export const yearSchema = z
+  .string({ message: "Parameter 'year' is required" })
```
laggycomputer, line 17: Did you actually intend this message to trigger on both an absent and an incorrectly typed input? Remember that, for example, on GraphQL, an incoming `Int!` would fail to validate. Then implying the year is missing would be incorrect!

laggycomputer (changes requested):

.vscode/settings.json
laggycomputer, line 1: This should not be committed. IDE settings are a user preference and in principle should not be committed to source trees. In this specific case it's also empty.

laggycomputer (changes requested):

.vscode/settings.json
sagardhunna: Oh okay, I can just delete this file right?

apps/api/src/schema/lib/year.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from "@hono/zod-openapi";
+
+/**
+ * Expects a string forming a four-digit positive integer
+ */
+
+export const yearSchema = z.coerce
+  .string()
+  .refine((val) => val !== ("" || "undefined" || "null"), {
```
laggycomputer, line 9: Use `required_error` and `invalid_type_error`. This example is given on the Zod docs:
```ts
const name = z.string({
  required_error: "Name is required",
  invalid_type_error: "Name must be a string",
});
```

laggycomputer (changes requested):
Almost done.

laggycomputer (approved):

andrew-wang0 (approved):
Good work!

Issue #96, opened 2025-01-27
`/v2/rest/programs/specializations` should have some way to filter by major id


Awesome-E:
Currently, this endpoint just returns a list of every spec that exists.
<img width="470" alt="Image" src="https://github.com/user-attachments/assets/c2a71b1e-f616-4531-b752-39af9216db72" />

For someone who might want to get specializations for a CS major, for example, It would be a more ideal flow to...
1. The user gets `/programs/majors` to get the major ID
2. The user gets `/programs/specializations?majorID=BS-201` and it would return just the CS Specs.

PR #97, opened 2025-01-28
feat: add way to filter `programs` endpoints by `majorid` and `id`

waterkimchi:
## Description

Previously, the endpoints for programs returned a comprehensive list of all `majors`, `minors`, and `specializations` without any filtering options. With this update, we have added the ability to query specific data by including optional parameters like `majorId` or `minorId`. This enhancement allows users to request only the relevant `specializations` associated with a particular `major` or `minor`, streamlining the data retrieval process and improving the overall user experience.

## Related Issue

`/v2/rest/programs/specializations` should have some way to filter by major id - #96 

## Motivation and Context

Currently, the endpoint returns a list of all available specializations, which can be overwhelming and inefficient, especially for users seeking information about a specific `major`. By implementing this fix, users will be able to first retrieve the `major ID` through `/programs/majors`, and then query `/programs/specializations?majorID=BS-201` to get only the relevant specializations for a particular major, such as `Computer Science`.

## How Has This Been Tested?

Tested using local environment.

REST:

Majors:
Screenshot 2025-01-27 at 21 25 04

Minors:
Screenshot 2025-01-27 at 21 25 17

Specializations:
Screenshot 2025-01-27 at 21 25 34

GraphQL:

Majors:
Screenshot 2025-01-27 at 23 14 26

Minors:
Screenshot 2025-01-27 at 22 07 38

Specializations:
Screenshot 2025-01-27 at 22 09 06

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/graphql/resolvers/programs.ts
+++ b/apps/api/src/graphql/resolvers/programs.ts
@@ -1,7 +1,10 @@
 import type { GraphQLContext } from "$graphql/graphql-context";
 import {
+  majorQuerySchema,
   majorRequirementsQuerySchema,
+  minorQuerySchema,
   minorRequirementsQuerySchema,
+  specializationQuerySchema,
   specializationRequirementsQuerySchema,
 } from "$schema";
 import { ProgramsService } from "$services";
@@ -39,27 +42,30 @@ export const programResolvers = {
         });
       return res;
     },
-    majors: async (_: unknown, args: unknown, { db }: GraphQLContext) => {
+    majors: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
+      const parsedArgs = majorQuerySchema.parse(args?.query);
       const service = new ProgramsService(db);
-      const res = await service.getMajors();
+      const res = await service.getMajors(parsedArgs);
       if (!res)
         throw new GraphQLError("Major data not found", {
           extensions: { code: "NOT_FOUND" },
         });
       return res;
     },
-    minors: async (_: unknown, args: unknown, { db }: GraphQLContext) => {
+    minors: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
+      const parsedArgs = minorQuerySchema.parse(args?.query);
       const service = new ProgramsService(db);
-      const res = await service.getMinors();
+      const res = await service.getMinors(parsedArgs);
       if (!res)
         throw new GraphQLError("Minor data not found", {
           extensions: { code: "NOT_FOUND" },
         });
       return res;
     },
-    specializations: async (_: unknown, args: unknown, { db }: GraphQLContext) => {
+    specializations: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
+      const parsedArgs = specializationQuerySchema.parse(args?.query);
       const service = new ProgramsService(db);
-      const res = await service.getSpecializations();
+      const res = await service.getSpecializations(parsedArgs);
       if (!res)
         throw new GraphQLError("Specializations data not found", {
           extensions: { code: "NOT_FOUND" },
--- a/apps/api/src/graphql/resolvers/programs.ts
+++ b/apps/api/src/graphql/resolvers/programs.ts
@@ -1,11 +1,11 @@
 import type { GraphQLContext } from "$graphql/graphql-context";
 import {
-  majorQuerySchema,
   majorRequirementsQuerySchema,
-  minorQuerySchema,
+  majorsQuerySchema,
   minorRequirementsQuerySchema,
-  specializationQuerySchema,
+  minorsQuerySchema,
   specializationRequirementsQuerySchema,
+  specializationsQuerySchema,
 } from "$schema";
 import { ProgramsService } from "$services";
 import { GraphQLError } from "graphql/error";
@@ -43,7 +43,7 @@ export const programResolvers = {
       return res;
     },
     majors: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
-      const parsedArgs = majorQuerySchema.parse(args?.query);
+      const parsedArgs = majorsQuerySchema.parse(args?.query);
       const service = new ProgramsService(db);
       const res = await service.getMajors(parsedArgs);
       if (!res)
@@ -53,7 +53,7 @@ export const programResolvers = {
       return res;
     },
     minors: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
-      const parsedArgs = minorQuerySchema.parse(args?.query);
+      const parsedArgs = minorsQuerySchema.parse(args?.query);
       const service = new ProgramsService(db);
       const res = await service.getMinors(parsedArgs);
       if (!res)
@@ -63,7 +63,7 @@ export const programResolvers = {
       return res;
     },
     specializations: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
-      const parsedArgs = specializationQuerySchema.parse(args?.query);
+      const parsedArgs = specializationsQuerySchema.parse(args?.query);
       const service = new ProgramsService(db);
       const res = await service.getSpecializations(parsedArgs);
       if (!res)
--- a/apps/api/src/graphql/schema/programs.ts
+++ b/apps/api/src/graphql/schema/programs.ts
@@ -66,10 +66,18 @@ input ProgramRequirementsQuery {
     programId: String!
 }
 
+input MajorIdQuery {
+    majorId: String!
+}
+
+input MinorIdQuery {
+    id: String!
+}
+
 extend type Query {
-    majors: [MajorPreview!]!
-    minors: [MinorPreview!]!
-    specializations: [SpecializationPreview!]!
+    majors(query: MajorIdQuery): [MajorPreview!]!
+    minors(query: MinorIdQuery): [MinorPreview!]!
+    specializations(query: MajorIdQuery): [SpecializationPreview!]!
     major(query: ProgramRequirementsQuery!): Program!
     minor(query: ProgramRequirementsQuery!): Program!
     specialization(query: ProgramRequirementsQuery!): Program!
--- a/apps/api/src/graphql/schema/programs.ts
+++ b/apps/api/src/graphql/schema/programs.ts
@@ -66,22 +66,22 @@ input ProgramRequirementsQuery {
     programId: String!
 }
 
-input majorsQuery {
+input MajorsQuery {
     id: String!
 }
 
-input minorsQuery {
+input MinorsQuery {
     id: String!
 }
 
-input specializationsQuery {
+input SpecializationsQuery {
     majorId: String!
 }
 
 extend type Query {
-    majors(query: majorsQuery): [MajorPreview!]!
-    minors(query: minorsQuery): [MinorPreview!]!
-    specializations(query: specializationsQuery): [SpecializationPreview!]!
+    majors(query: MajorsQuery): [MajorPreview!]!
+    minors(query: MinorsQuery): [MinorPreview!]!
+    specializations(query: SpecializationsQuery): [SpecializationPreview!]!
     major(query: ProgramRequirementsQuery!): Program!
     minor(query: ProgramRequirementsQuery!): Program!
     specialization(query: ProgramRequirementsQuery!): Program!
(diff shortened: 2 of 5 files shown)


laggycomputer:
This is probably a feature update, not a bugfix, and is missing GraphQL.

apps/api/src/schema/programs.ts
```diff
@@ -1,6 +1,28 @@
 import { z } from "@hono/zod-openapi";
 
 const programIdBase = z.string({ required_error: "programId is required" });
+const majorIdBase = z.string();
+
+export const majorQuerySchema = z.object({
+  majorId: majorIdBase.optional().openapi({
+    description: "A major ID to query majors for",
```
laggycomputer, line 8: This is a bit clunky; the meaning you're trying to convey is "if you specify an ID, you'll only get that major back". Saying "major" twice is a bit awkward.

apps/api/src/schema/programs.ts
```diff
@@ -1,6 +1,28 @@
...
+    example: "BA-163",
+  }),
+});
+
+export const minorQuerySchema = z.object({
+  id: z.string().optional().openapi({
+    description: "A minor ID to query minors for",
```
laggycomputer, line 15: Same here as above.

apps/api/src/schema/programs.ts
```diff
@@ -1,6 +1,28 @@
...
+    example: "49A",
+  }),
+});
+
+export const specializationQuerySchema = z.object({
+  majorId: majorIdBase.optional().openapi({
+    description: "A major ID to query specializations for",
```
laggycomputer, line 22: You should make it clear how majors and specializations are associated for this description: the idea is "if you specify a major, only specializations for that major will be returned".

apps/api/src/services/programs.ts
```diff
@@ -11,7 +14,23 @@ import type { z } from "zod";
 export class ProgramsService {
   constructor(private readonly db: ReturnType<typeof database>) {}
 
-  async getMajors() {
+  async getMajors(query: z.infer<typeof majorQuerySchema>) {
```
laggycomputer, line 17: Why do an early return and not a conditional `where`? It may help to know that a `.where` with no arguments does not add any conditions. Having two code paths with such similar code is an extra maintenance burden and hurts readability, since early return is usually used to checking preconditions or special cases, not a full code path even more complex than the "main" path outside the early return.

laggycomputer (changes requested):

apps/api/src/graphql/schema/programs.ts
```diff
@@ -66,10 +66,18 @@ input ProgramRequirementsQuery {
...
+
+input MinorIdQuery {
+    id: String!
+}
+
 extend type Query {
-    majors: [MajorPreview!]!
-    minors: [MinorPreview!]!
-    specializations: [SpecializationPreview!]!
+    majors(query: MajorIdQuery): [MajorPreview!]!
```
laggycomputer, line 78: Existing naming convention suggests these should be named `MajorsQuery`, etc.

apps/api/src/graphql/schema/programs.ts
```diff
@@ -66,10 +66,18 @@ input ProgramRequirementsQuery {
...
+    minors(query: MinorIdQuery): [MinorPreview!]!
+    specializations(query: MajorIdQuery): [SpecializationPreview!]!
```
laggycomputer, line 80: Even if the field is identical in name and type, it does carry a different meaning when used for the specializations query. I argue it should be a distinct type.

waterkimchi:
I apologize for the dirty code. Thank you for all the reviews.

apps/api/src/schema/programs.ts
```diff
@@ -1,6 +1,28 @@
 import { z } from "@hono/zod-openapi";
 
 const programIdBase = z.string({ required_error: "programId is required" });
+const majorIdBase = z.string();
+
+export const majorQuerySchema = z.object({
+  majorId: majorIdBase.optional().openapi({
+    description: "The ID of the specific major being requested for",
```
laggycomputer, line 8: "being requested for" is awkward, could be more concise: "The ID of a single major to request, if provided" or something along those lines is most concise.

apps/api/src/schema/programs.ts
```diff
@@ -1,6 +1,28 @@
...
+export const minorQuerySchema = z.object({
+  id: z.string().optional().openapi({
+    description: "The ID of the specific minor being requested for",
+    example: "49A",
+  }),
+});
+
+export const specializationQuerySchema = z.object({
+  majorId: majorIdBase.optional().openapi({
+    description: "The ID of a major to query its associated specializations",
```
laggycomputer, line 22: Also a bit strange, try e.g. "Only fetch specializations associated with the major with this ID, if provided"

apps/api/src/services/programs.ts
```diff
@@ -36,25 +39,43 @@ export class ProgramsService {
       .from(majorSpecialization)
       .innerJoin(major, eq(majorSpecialization.id, major.id))
       .innerJoin(degree, eq(major.degreeId, degree.id));
+
+    if (query.majorId) {
+      majorBuilder.where(eq(major.id, query.majorId));
```
laggycomputer, line 44: Queries are immutable, meaning this line creates a new query with your `.where` and immediately drops it. The solution is to instead do an unconditional `.where` and, inside, spread the result of a ternary which checks the query. In code, something like:
`.where(...query.majorId ? [eq(major.id, query.majorId)] : [])`

Do not use `let` and reassign; it is generally bad practice to introduce `let` unnecessarily.

apps/api/src/services/programs.ts
```diff
@@ -36,25 +39,43 @@ export class ProgramsService {
...
+    }
+
+    return majorBuilder;
   }
 
-  async getMinors() {
-    return this.db
+  async getMinors(query: z.infer<typeof minorQuerySchema>) {
+    const minorBuilder = this.db
```
laggycomputer, line 51: Same here as above. Note that if you included screenshots of these query parameters in action or found another way to hold yourself accountable to test this new feature, this bug would have been caught.

laggycomputer (changes requested):

laggycomputer:
> I apologize for the dirty code. Thank you for all the reviews.

You're good, first time for everything. Thanks for getting on it so fast.

apps/api/src/graphql/schema/programs.ts
```diff
@@ -66,10 +66,18 @@ input ProgramRequirementsQuery {
     programId: String!
 }
 
+input MajorIdQuery {
+    majorId: String!
```
laggycomputer, line 70: If you split this type as I suggest below, then the field for the majors query should be named just `id`.

laggycomputer (changes requested):

apps/api/src/graphql/resolvers/programs.ts
```diff
@@ -1,7 +1,10 @@
 import type { GraphQLContext } from "$graphql/graphql-context";
 import {
+  majorQuerySchema,
```
laggycomputer, line 3: These should be named in the plural, `majorsQuerySchema`, etc.

apps/api/src/graphql/schema/programs.ts
```diff
@@ -66,10 +66,14 @@ input ProgramRequirementsQuery {
     programId: String!
 }
 
+input ProgramQuery {
+    id: String!
```
laggycomputer, line 70: I argued earlier that this should actually be three different types, and I stand by that now. By combining them into one type you lost even more meaning about the semantics of the `id` field, which really does behave differently among these three resources.

apps/api/src/schema/programs.ts
```diff
@@ -2,6 +2,27 @@ import { z } from "@hono/zod-openapi";
...
+export const majorQuerySchema = z.object({
+  id: z.string().optional().openapi({
+    description: "The ID of a single major to request, if provided",
+    example: "BA-163",
+  }),
+});
+
+export const minorQuerySchema = z.object({
+  id: z.string().optional().openapi({
+    description: "The ID of the specific minor being requested for",
```
laggycomputer, line 14: This should be reworded consistent with the description for the equivalent field on the majors query.

laggycomputer (changes requested):

laggycomputer:
Good use of SQL literals, by the way.

apps/api/src/graphql/schema/programs.ts
```diff
@@ -66,10 +66,22 @@ input ProgramRequirementsQuery {
     programId: String!
 }
 
+input majorsQuery {
```
laggycomputer, line 69: Very silly of me to not tell you this earlier but types in GraphQL are in `PascalCase`, so `MajorsQuery`

laggycomputer (changes requested):
One little thing, in the meantime I'll pull down for manual testing.

laggycomputer (approved):
This worked locally before your latest change. Appreciate the speed.

apps/api/src/services/programs.ts
```diff
@@ -34,27 +37,30 @@ export class ProgramsService {
         division: degree.division,
       })
       .from(majorSpecialization)
+      .where(query.id ? eq(majorSpecialization.id, query.id) : sql`true`)
```
andrew-wang0, line 40: ```sql'true'``` is a valid solution here, but we'll want to use `undefined` since it better denotes that we don't  want to execute a where clause (rather than executing WHERE TRUE)

```suggestion
      .where(query.id ? eq(majorSpecialization.id, query.id) : undefined)
```

apps/api/src/services/programs.ts
```diff
@@ -34,27 +37,30 @@ export class ProgramsService {
...
-  async getMinors() {
+  async getMinors(query: z.infer<typeof minorsQuerySchema>) {
     return this.db
       .select({
         id: minor.id,
         name: minor.name,
       })
-      .from(minor);
+      .from(minor)
+      .where(query.id ? eq(minor.id, query.id) : sql`true`);
```
andrew-wang0, line 52: Same here

apps/api/src/services/programs.ts
```diff
@@ -34,27 +37,30 @@ export class ProgramsService {
...
+  async getSpecializations(query: z.infer<typeof specializationsQuerySchema>) {
     return this.db
       .select({
         id: specialization.id,
         majorId: specialization.majorId,
         name: specialization.name,
       })
-      .from(specialization);
+      .from(specialization)
+      .where(query.majorId ? eq(specialization.majorId, query.majorId) : sql`true`);
```
andrew-wang0, line 63: Same here

andrew-wang0 (commented):
@waterkimchi Great work on this :raised_hands:, just a small semantic fix on the sql queries

andrew-wang0:
@laggycomputer it looks like the majors route is only returning majors with specializations from the `innerJoin` on `major_specialization`. Can you confirm?

apps/api/src/services/programs.ts
waterkimchi: Tried to use `undefined` in [8905d30](https://github.com/icssc/anteater-api/pull/97/commits/8905d30290dfad72a96c33d81739ac2404aa37bd) which resulted in a staging failure which was resolved with [90749e6](https://github.com/icssc/anteater-api/pull/97/commits/90749e698c70dad0533bef934b2428ffae85bfe1)

laggycomputer:
Sounds right and also sounds really silly on my part.

apps/api/src/services/programs.ts
andrew-wang0: Hmmm that's odd, let me do some digging

andrew-wang0: > Tried to use `undefined` in [8905d30](https://github.com/icssc/anteater-api/pull/97/commits/8905d30290dfad72a96c33d81739ac2404aa37bd) which resulted in a staging failure. Resolved with [90749e6](https://github.com/icssc/anteater-api/pull/97/commits/90749e698c70dad0533bef934b2428ffae85bfe1).

@waterkimchi looks like it may have been a one time issue with the github action, [90b34d2](https://github.com/icssc/anteater-api/pull/97/commits/90b34d2d53a2e46ed72bc31156f7348a5cfbce74) staged without error. If you come across this again, feel free to lmk to rerun the action.

andrew-wang0 (approved):
:fire:

Issue #19, opened 2024-11-03
[Courses/Instructors] Cursor based pagination


ecxyzzy:
## Expected Behavior
The courses and instructors routes should also support cursor-based pagination.

## Current Behavior
The courses and instructors routes only support skip-and-take pagination, which is less efficient for queries where the desired item(s) is/are later in the result set.

PR #101, opened 2025-01-28
feat: courses & instructor cursor based pagination

estao1:
## Description

Added cursor-based pagination using unique identifiers (`id` for courses and `ucinetid` for instructors) as an more efficient alternative to skip-take pagination. Includes new `/coursesCursor` and `/instructorsCursor` routes with `cursor` and `take` parameters, which returns paginated results with a `nextCursor` field.

## Related Issue

Completes #19.

## Motivation and Context

Skip-and-take pagination can become less efficient for queries where the desired items are later in the set. Cursor-based pagination helps ensure consistent query performance, especially for larger datasets.

## How Has This Been Tested?

Tested changes locally on API docs and Yoga GraphiQL.

image
image
image
image
image
image

## Types of changes

## Checklist:


laggycomputer:
I get a 500 on GraphQL introspection. Can you confirm?

estao1:
Somehow forgot to commit this fix, should be good now.

apps/api/src/rest/routes/courses.ts
```diff
@@ -11,10 +12,11 @@ import {
   responseSchema,
 } from "$schema";
 import { CoursesService } from "$services";
-import { OpenAPIHono, createRoute } from "@hono/zod-openapi";
+import { OpenAPIHono, createRoute, z } from "@hono/zod-openapi";
```
laggycomputer, line 15: It has not been the norm to import and declare these schemas inline to REST; they are usually defined in the relevant schemas file and imported to REST. I understand it's simple here but in the future we may want, say, GraphQL to parse with this schema. This would mean GraphQL would import and depend on REST code, which doesn't make sense.

apps/api/src/rest/index.ts
```diff
@@ -17,9 +17,11 @@ const restRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
 
 restRouter.route("/calendar", calendarRouter);
 restRouter.route("/courses", coursesRouter);
+restRouter.route("/coursesCursor", coursesCursorRouter);
```
laggycomputer, line 20: What are your thoughts on using `/courses/cursor` and `/instructors/cursor` or `/instructors/byCursor/`? I'm interested in minimizing the number of root endpoints.

apps/api/src/rest/routes/courses.ts
```diff
@@ -99,6 +101,39 @@ const coursesByFiltersRoute = createRoute({
   },
 });
 
+const coursesByCursorRoute = createRoute({
+  summary: "Filter courses by cursor",
```
laggycomputer, line 105: "by cursor" doesn't quite make sense; try coping the existing `coursesRoute` summary and just add "by cursor" to indicate you're providing a cursor into the results.

apps/api/src/schema/courses.ts
```diff
@@ -75,6 +75,32 @@ export const coursesQuerySchema = z.object({
...
+  minUnits: z.coerce.number().optional(),
+  maxUnits: z.coerce.number().optional(),
+  descriptionContains: z.string().optional(),
+  geCategory: z
+    .enum(inputGECategories, {
+      message:
+        "If provided, 'geCategory' must be one of 'GE-1A', 'GE-1B', 'GE-2', 'GE-3', 'GE-4', 'GE-5A', 'GE-5B', 'GE-6', 'GE-7', or 'GE-8'",
+    })
+    .optional(),
+  cursor: z.string().optional(),
```
laggycomputer, line 97: You should leave a description clearly referencing the cursor query parameter. (repeat for the other cursor)

apps/api/src/schema/courses.ts
```diff
@@ -75,6 +75,32 @@ export const coursesQuerySchema = z.object({
...
+  take: z.coerce
+    .number()
+    .default(100)
+    .refine((x) => x <= 100, "Page size must be smaller than 100"),
```
laggycomputer, line 101: Your error here doesn't make sense for `x === 100`. (repeat for the other cursor)

laggycomputer (changes requested):

apps/api/src/graphql/resolvers/courses.ts
```diff
@@ -20,5 +20,15 @@ export const coursesResolvers = {
       const service = new CoursesService(db);
       return await service.batchGetCourses(ids);
     },
+    coursesByCursor: async (_: unknown, args: { query: unknown }, { db }: GraphQLContext) => {
```
laggycomputer, line 23: Try to keep the order of the resolvers in line with the order of the underlying functions on the service.

apps/api/src/graphql/schema/courses.ts
```diff
@@ -49,9 +49,29 @@ input CoursesQuery {
...
+type CoursesByCursor {
+    items: [Course!]!
+    nextCursor: String
+}
+
 extend type Query {
     course(id: String!): Course!
     courses(query: CoursesQuery!): [Course!]!
     batchCourses(ids: [String!]!): [Course!]!
+    coursesByCursor(query: CoursesByCursorQuery!): CoursesByCursor!
```
laggycomputer, line 75: Same here, try to order everything consistently everywhere.

apps/api/src/rest/routes/courses.ts
```diff
@@ -11,10 +12,11 @@ import {
   responseSchema,
 } from "$schema";
 import { CoursesService } from "$services";
-import { OpenAPIHono, createRoute } from "@hono/zod-openapi";
+import { OpenAPIHono, createRoute, z } from "@hono/zod-openapi";
 import { database } from "@packages/db";
 
 const coursesRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
+const coursesCursorRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
```
laggycomputer, line 19: What was your thought process to create a new router? I don't see a reason myself but you're free to add context. (same for instructors)

laggycomputer (changes requested):

apps/api/src/rest/routes/courses.ts
andrew-wang0: I can jump in here. Originally we wanted to have it in the same route, but we decided that having the response schema change based on a query parameter would introduce too much complexity.

laggycomputer: Got it, forgot about that aspect of the architecture. Thanks.

apps/api/src/schema/courses.ts
```diff
@@ -75,6 +75,32 @@ export const coursesQuerySchema = z.object({
...
+      message: "If provided, 'courseLevel' must be 'LowerDiv', 'UpperDiv', or 'Graduate'",
+    })
+    .optional(),
+  minUnits: z.coerce.number().optional(),
+  maxUnits: z.coerce.number().optional(),
+  descriptionContains: z.string().optional(),
+  geCategory: z
+    .enum(inputGECategories, {
+      message:
+        "If provided, 'geCategory' must be one of 'GE-1A', 'GE-1B', 'GE-2', 'GE-3', 'GE-4', 'GE-5A', 'GE-5B', 'GE-6', 'GE-7', or 'GE-8'",
```
laggycomputer, line 94: There's no need to list the enum options; the API docs already do that.

laggycomputer, line 94: There's no need to list enum variants; the docs do this for you automatically.

laggycomputer (changes requested):

apps/api/src/rest/index.ts
estao1: I originally tried that route, but the route `/courses/{id}` conflicted with `/courses/cursor`, leading to the query trying to find a course with `id=cursor`. Andrew and I decided the best way around this was to name the routes `/coursesCursor` and `/instructorsCursor`.

apps/api/src/rest/routes/courses.ts
estao1: Not sure if I understand. The courses without cursor route's summary is just "Filter courses", so isn't what I have the same but with "by cursor" added?

laggycomputer (changes requested):

apps/api/src/schema/search.ts
```diff
@@ -7,7 +7,7 @@ export const searchQuerySchema = z.object({
   take: z.coerce
     .number()
     .default(100)
-    .refine((x) => x <= 100, "Page size must be smaller than 100"),
+    .refine((x) => x <= 100, "Page size must be less than or equal to 100"),
```
laggycomputer, line 10: There is a `.gte` built into Zod, which does what you want more concisely. It is equivalent and you do not have to use it; the current form is clear enough.

laggycomputer (changes requested):

laggycomputer (approved):

laggycomputer:
There's also `.int`; you might want to use that for validation.

andrew-wang0 (approved):
:fire:

Issue #18, opened 2024-10-31
[WebSoc] Related courses from section code range


ecxyzzy:
Feature request from @alexanderl19 discussed over call.

## Expected Behavior
When this flag is enabled and a valid range of section codes is provided, should do the following:
- Find all courses where at least one section is in this range.
- Return a standard WebSoc response containing all of these courses and related metadata, as if they had been queried normally.

PR #104, opened 2025-01-31
feat(websoc): implement related courses from section code range

aadi-shanker:
## Description
- Add includeRelatedCourses flag to the WebSoc query schema (coerced from string to boolean).
- Updated the `buildQuery` function in `services/websoc.ts` to build conditions for related courses based on section code ranges. When includeRelatedCourses is enabled along with `sectionCodes`, a subquery fetches the related course IDs, and the query is constructed to filter based on these IDs.

## Related Issue
Completes #18 

## Motivation and Context
This change enhances the WebSoc API by providing a more flexible and user-friendly way to fetch related courses. The `includeRelatedCourses` flag now supports multiple boolean representations and ensures that related courses are dynamically fetched based on section code ranges. 

## How Has This Been Tested?
Manually tested the REST and GraphQL endpoints using various scenarios:
- Valid queries using only `sectionCodes` with `includeRelatedCourses`: `"true"`.
- Queries using only `courseIds`.
- Queries that combine both `sectionCodes` and `courseIds` with `includeRelatedCourses`, which correctly trigger the error response.

image

## Types of changes

## Checklist:


apps/api/src/graphql/schema/websoc.ts
```diff
@@ -110,6 +110,7 @@ input WebsocQuery {
     startTime: String
     endTime: String
     excludeRestrictionCodes: String
+    includeRelatedCourses: String
```
laggycomputer, line 113: Since GraphQL supports a real `Boolean` type, you should use that.

apps/api/src/schema/websoc.ts
```diff
@@ -155,6 +155,24 @@ export const websocQuerySchema = z.object({
       }
       return parsedCodes;
     }),
+  includeRelatedCourses: z
```
laggycomputer, line 158: And given the above you need to coerce this string...

apps/api/src/schema/websoc.ts
```diff
@@ -155,6 +155,24 @@ export const websocQuerySchema = z.object({
...
+    .string()
+    .optional()
+    .transform((value, ctx) => {
+      if (!value) return false;
```
laggycomputer, line 162: And check for `"undefined"` in this function (I think that's how `.coerce` works...)

apps/api/src/schema/websoc.ts
```diff
@@ -155,6 +155,24 @@ export const websocQuerySchema = z.object({
...
+      }
+      if (normalizedValue === "false" || normalizedValue === "0" || normalizedValue === "no") {
+        return false;
+      }
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "Invalid value for includeRelatedCourses. Must be true/false, 0/1, or yes/no.",
+      });
+      return z.NEVER;
+    }),
```
laggycomputer, line 175: Missing documentation; what is a "related course"? Are there any conditions under which this parameter is ineffective (i.e. if no `sectionCodes` are given)? What is an example?

apps/api/src/services/websoc.ts
```diff
@@ -91,6 +91,22 @@ function buildQuery(input: WebsocServiceInput) {
     }
     conditions.push(or(...courseNumberConditions));
   }
+  if (input.includeRelatedCourses && input.sectionCodes) {
```
laggycomputer, line 94: This `if` is identical to the one below it except for the additional check on `input.includeRelatedCourses`. Without documentation I don't understand your original intent; was this a typo? If not, what was your intended meaning?

laggycomputer (changes requested):

laggycomputer:
Just confirmed; this PR as is does not fulfill the written feature request from the issue. Remember the flag was meant to set the following behavior:
If a section code filter is provided, and a section with a certain `code` is returned, then every other section from the same course must also be included. For example, this query is not behaving correctly:
```graphql
query Websoc {
    websoc(
        query: {
            includeRelatedCourses: "true"
            year: "2025"
            quarter: Winter
            sectionCodes: "35610"
        }
    ) {
        schools {
            departments {
                courses {
                    courseNumber
                    sections {
                        sectionType
                        sectionCode
                    }
                }
            }
        }
    }
}
```
We would expect it to return every ICS 33 section (including the other lecture and the labs) this term, since ICS 33 is a "related course" to a section within the set of codes we specified. However it only returns the section we originally requested.

apps/api/src/services/websoc.ts
aadi-shanker: Yes, I misunderstood the purpose of the includeRelatedCourses flag. Initially, I thought the flag was meant to return only the sections provided in the input, which is already implemented in `buildQuery`. I am currently working on updating the implementation to ensure that when section codes are provided and the flag is activated, all sections related to the same course (including other lectures, labs, and discussions) are returned. I will do this using `courseId` to identify and include those related sections.

ecxyzzy:
Also, please make sure to update the title of your PR to comply with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).

laggycomputer (changes requested):
I'm not sure why the new `courseIds` parameter is necessary. It's also not documented so a user would have no idea what it does or that it's only associated with related courses (and not meaningful otherwise) and mutually exclusive with `sectionCodes`. Remember that the spec says that you should be able to accomplish related courses without any additional parameters (or forced lack thereof).

You also accumulate SQL data into JS and then use it to request SQL again, which usually incurs excess computing, memory, and network cost. It can and should be done purely in SQL. Off the top of my head a self join or subquery can be effective.

aadi-shanker:
> I'm not sure why the new `courseIds` parameter is necessary. It duplicates existing parameters and isn't documented so a user would have no idea what it does differently from the existing options or that it's only associated with related courses (and not meaningful otherwise) and mutually exclusive with `sectionCodes`. Remember that the spec says that you should be able to accomplish related courses without any additional parameters (or forced lack thereof).
> 
> You also accumulate SQL data into JS and then use it to request SQL again, which usually incurs excess computing, memory, and network cost. It can and should be done purely in SQL. Off the top of my head a self join or subquery can be effective.

Hey Dante,

Thanks for the feedback. I introduced the courseIds parameter based on Andrew’s guidance. The intent is twofold:

- It allows users to directly query for a list of course IDs (e.g. &courseIds=WOMNST187,WOMNST50C,WOMNST60B), so they can explicitly request all sections for specific courses.

- It separates the concerns of query construction and execution—when sectionCodes are provided with includeRelatedCourses, a subquery is used to fetch the related course IDs and then those are passed into the main query.

I realize that in the current implementation the courseIds value gets overwritten when both sectionCodes and `includeRelatedCourses` are provided, which might seem confusing from a user’s perspective. Taking that into account i asked Andrew what would be the best approach and he said to either have impelementation to merge the courseIds or just let the user know `includeRelatedCoures` only works with section codes. I'll make sure to update the documentation to clarify that the courseIds parameter is intended for direct queries, and that when using includeRelatedCourses, only sectionCodes should be supplied.

Regarding the SQL data accumulation in JS before making another query, I understand the performance concern. This design was intentional for clear separation of query building and execution, but I'm open to exploring a more SQL-centric solution so I can refactor the implementation.

Let me know if this approach works or if further adjustments are needed.

laggycomputer:
Thanks for the clarification.
I'd like to hear more of your approach and broader context, if you don't mind.
Doing so would allow me to deeply understand and comment directly on the code.

> It allows users to directly query for a list of course IDs

What about existing `department` and `courseNumber` parameters?
It seems dissonant to have two ways to ask for sections by course.
What have you and Andrew decided to do about this?

> when sectionCodes are provided with includeRelatedCourses, a subquery is used to fetch the related course IDs and then those are passed into the main query

From what I see so far, the first DB query which is de-duplicated via `Set` (assuming that's what you meant) is not a sub-query.
A sub-query is a specific concept in SQL where a `SELECT` statement appears in place of a reference to a table in a `JOIN`, `SELECT FROM`, etc.

> It separates the concerns of query construction and execution
I agree that this is a good priority to have.
I also think there may be a way to both have the cake and eat it by building conditions which also include a SQL sub-query or CTE when necessary.
I argue that this would still be an acceptable separation of concerns since that expression isn't executed by the query building code.

> either have impelementation to merge the courseIds or just let the user know includeRelatedCoures only works with section codes
The latter seems reasonable; to be clear I understand you want to make the inclusion of the `sectionCodes` parameter necessary and sufficient to allow `includeRelatedCoures` to be effective.
I suggest you extend this to any query which does not include a filter by course or department, since I can imagine use cases where this might be useful e.g. asking "what are the sections of classes which have a section taught by this instructor?"

But "impelementation to merge the courseIds" confuses me; can you clarify?

> Could you please share any specific suggestions or examples of how you'd like to see this implemented?
Of course.
At a high level, you can reframe the `buildQuery` function into two broad cases:

First, when related courses functionality is not required; this case is the existing code from before this PR.

Second, when related courses functionality *is* needed; this is the new case.
You have the choice of when this case should trigger (see above).
In this second case, you should wrap the conditions from case 1 in a sub-query, CTE, etc.
This inner expression should fetch according to the conditions from case 1, but `SELECT` only the `course_id` column.
Note that this is different from your usage of the term "course ID" so far where you mean the human readable department and number.
The `course_id` column is a UUID.
Finally, the return value of `buildQuery` in this case is simply the sections whose `course_id` is among the `course_id`s from the sub-query/CTE.

Let me know if you have questions on any of the above.
Thanks for your detailed responses to my review.

laggycomputer:
> User queries with includeRelatedCourses → run a separate JavaScript DB query to find related course IDs (which now I realize not a true subquery) → store those courseIds in input.courseIds → pass input.courseIds into buildQuery → execute the final SQL query to find all sections of those related courses.

That agrees with my understanding.

> the expectation is to handle this logic within the main SQL query itself, rather than fetching courseIds separately in JavaScript and passing them into buildQuery.

It should be in SQL, not necessarily the last query you do or any other query you might call the "main" query. But it should not cross the SQL-JS barrier and go back in again without a very good reason (e.g. you've proven it's actually faster).

> Given that, I’ll be looking into refactoring this into a fully SQL-based solution.

Excellent.

> I’ll need to determine how to prioritize those conditions in buildQuery. Let me know if you have any preferences on that!

They are not mutually exclusive. Of course we are working under the SemVer constraint that we will have no breaking changes. If they are present together, I argue disjunction (logical OR) is most appropriate.

> I am also assuming this problem would also change due to the change in subquery implementation that I need to refactor.

Your decision to disallow the presence of all three parameters is mostly a concern of your validation step. Your further code, including SQL queries, can assume that invariant is upheld where applicable. As particularly applied to SQL, it would probably \*reduce* the amount of cases you have to consider when building queries.

> Should the subquery (or CTE) be built using Drizzle’s TypeScript DSL, or would raw SQL be preferred for this part?

Prefer the ORM, since we have it. As I understand the problem statement this should be feasible unless you discover yet another uncovered corner case in Drizzle. If you have to use raw SQL, minimize its scope (e.g. if you can't express a single function, use raw SQL for only that function).

> I just want to confirm that this aligns with your expectations and that I’m understanding the trade-offs correctly.

You should never attempt a join on the human-readable course IDs unless you have a very very good reason. You should also never expose the UUIDs. Your understanding is correct.

> Are there any specific examples or patterns from our codebase that you think would serve as a good reference for implementing this subquery or CTE approach?

A CTE is used [here](https://github.com/icssc/anteater-api/blob/77dbd87f932a57da54a0514c77faa2450ea932ec/apps/api/src/services/programs.ts#L20) to assemble an array of specializations for each major. Here, we `SELECT` from a CTE.

Two sub-queries are used [here](https://github.com/icssc/anteater-api/blob/c20ae6102bf68c551e35ec413573288de6c6d47e/apps/api/src/services/enrollment-changes.ts#L112). The first is `JOIN`ed against to accomplish a complex filtering operation, forming the second. The second is `SELECT`ed against to do a more simple filtering operation. Don't worry about the complexity, just understand what methods need to be called. After you create a CTE or sub-query, you can use it in place of a table identifier (mostly).

> I wanted to give you a heads-up that I should have the new implementation ready within the next week or so. I have my ICS 51 midterm this Wednesday, so I’m pretty packed with studying until then.

No worries. I'm told this isn't urgent.

apps/api/src/services/websoc.ts
```diff
@@ -387,18 +387,23 @@ function transformTerm(term: { year: string; quarter: Term }) {
...
-        school: getTableColumns(websocSchool),
-        department: getTableColumns(websocDepartment),
-        course: getTableColumns(websocCourse),
-        section: getTableColumns(websocSection),
-      })
+  makeSelect(
+    selection: Parameters<ReturnType<typeof database>["select"]>[0],
+    includeFilterTables?: boolean,
+  ) {
+    // makeSelect<T extends SelectedFields>(selection: T, includeFilterTables?: boolean) {
```
sanskarm7, line 394: Can delete comment if unnecessary

laggycomputer: oh sorry that's excess typescript footgunning left over

laggycomputer (approved):

sanskarm7 (approved):
lgtm

Issue #93, opened 2025-01-24
Add missing openapi query & response specifications to schemas


andrew-wang0:
Extension of #79 

Currently, many keys in our schemas have no good examples or descriptions. 

For example, the below is missing examples
![Image](https://github.com/user-attachments/assets/2a734c44-70c5-41b7-a15d-b8c1246b9522)

And the blow is missing descriptions
![Image](https://github.com/user-attachments/assets/87868280-98e6-4960-8983-6bc1cbfbb4f0)

Add openapi metadata to all routes with missing examples and descriptions to help users understand the query params and response schemas.

See the LARC endpoint for an example of a fully documented endpoint using openapi.

rsahyoun:
I can take this task if that works for you.

PR #111, opened 2025-02-06
feat(documentation): Add descriptions and examples to API Schema

rsahyoun:
## Description

Updates to the schema files by adding descriptions for the parameters and examples in the API specs.
Specifically, I am adding examples of what users might see when their API call goes through successfully.

## Related Issue

related to issue: #93 

## Motivation and Context

This change in relevant because it gives the user of the api clear description of the parameters for a specific api call as well as examples of what the api should return if the call was done correctly.

## How Has This Been Tested?

I was able to test each of the changes I made to the schema examples and parameter descriptions by running it on my local environment and making sure that all my changes went through properly and followed the style of the rest of the api that was already complete.

image

image

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/rest/routes/courses.ts
+++ b/apps/api/src/rest/routes/courses.ts
@@ -23,27 +23,28 @@ coursesRouter.openAPIRegistry.register("prereq", prerequisiteSchema);
 coursesRouter.openAPIRegistry.register("prereqTree", prerequisiteTreeSchema);
 
 const batchCoursesRoute = createRoute({
-  summary: "Retrieve courses with IDs",
+  summary: "Retrieve multiple courses",
   operationId: "batchCourses",
   tags: ["Courses"],
   method: "get",
   path: "/batch",
   request: { query: batchCoursesQuerySchema },
-  description: "Retrieves courses with the IDs provided",
+  description:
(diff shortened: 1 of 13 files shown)


laggycomputer (changes requested):
I will refrain from point feedback because that would result in roughly 100 or more comments. They already lag my system. I will begin generally:

1. Many schemas and descriptions are repeated. Some existing schemas are not being reused. For the former, they should be outlined. As a maintainer I prefer perfect outlining and concise code to a slightly better, more specialized docstring. For the latter, particularly `yearSchema`, use the shared definition.
2. Verbosity: A lot of these strings, particularly the descriptions of responses by error code, are too verbose. Specialization is nice but it's also cumbersome and doesn't add much value. The code itself and its standardized meaning is usually enough documentation. As stated, prefer borrowing another definition verbatim to having several more words which are already implied by context.
3. Some schemas have been inlined, e.g. the cursor schemas, when they should not be.
4. Some docstrings are written incorrectly to describe a response field vs. a query parameter or making an error in number or similar grammatical mistake.
5. Do not give `enum` in `.openapi`, this is redundant.
6. Do not give an example of a literal.

Please re-request a review from me when substantial progress has been made on these points.

ecxyzzy:
Closing as this PR has had no activity for the past two weeks.

Issue #110, opened 2025-02-06
[ETL/WebSoc] Bisect chunk on failed chunk-wise scrape


ecxyzzy:
When we perform a chunk-wise scrape of WebSoc, there is a possibility that sections will have been added to a chunk between the previous and current scraping runs. To mitigate this, the scraper uses chunk sizes with a 1% margin of error built in, but a more robust failsafe should be implemented in addition to the current margin.

A good solution seems to be retrying the current chunk on failure, by scraping the first half of the chunk followed by the second half, while of course respecting the existing rate limits. One bisection should be sufficient as section codes appear to be roughly uniformly distributed, and thus a bisection of an oversized chunk should almost always result in less than 900 sections per half, but a proof is currently lacking. Better solutions with justifications are naturally welcome.

PR #114, opened 2025-02-07
feat(websoc): bisect to handle section chunk overrun

laggycomputer:
## Related Issue

Execute recommendation in #110.

## Motivation and Context

Currently, when `libwebsoc-next` finds more than 900 sections, it immediately throws.
In the context of our current scraping strategy using chunks of the section space, a chunk thought to contain less than 900 sections may be found (and, in practice, has been found) to actually contain more than 900 sections, causing the scraping of that chunk to fail, crashing the scraper.
A later run would crash too, at the exact same point.
So a violation of our current assumptions is fatal.

## How Has This Been Tested?

Change `SECTIONS_PER_CHUNK` to something greater than 900 (the higher, the more often the failsafe will trigger).
This will simulate more than 900 sections being found in a chunk.

Observe in this printout as a chunk is found to be too large and bisected once, which is successful:
```
Chunk 50004-52814 failed (probably too large); bisecting and trying again...
Scraping chunk 50004-51409
Inserted 0 enrollment entries
Scraping chunk 51410-52814
Inserted 0 enrollment entries
```

Observe that this bisection is also recursive and therefore resilient under a pathological case:
```
Scraping chunk 01012-12105
Chunk 01012-12105 failed (probably too large); bisecting and trying again...
Scraping chunk 01012-06558
Chunk 01012-06558 failed (probably too large); bisecting and trying again...
Scraping chunk 01012-03785
Inserted 0 enrollment entries
Scraping chunk 03786-06558
Inserted 0 enrollment entries
Scraping chunk 06559-12105
Inserted 0 enrollment entries
```

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/websoc-scraper/src/lib.ts
+++ b/apps/data-pipeline/websoc-scraper/src/lib.ts
@@ -40,7 +40,7 @@ import { load } from "cheerio";
 
 /**
  * WebSoc allows us to scrape up to 900 sections per chunk.
- * This provides a 1% margin of error in case sections magically appear within ranges.
+ * This provides a 1% margin of error in case new sections appear within ranges (which has happened before).
  */
 const SECTIONS_PER_CHUNK = 891;
 
@@ -830,16 +830,9 @@ export async function scrapeTerm(
   } else {
     console.log("Performing chunk-wise scrape.");
     for (let i = 0; i < sectionCodeBounds.length; i += 2) {
-      const lower = sectionCodeBounds[i];
-      const upper = sectionCodeBounds[i + 1] ?? LAST_SECTION_CODE;
-      const sectionCodes = `${lower}-${upper}`;
-      console.log(`Scraping chunk ${sectionCodes}`);
-      const resp = await request(term, {
-        sectionCodes,
-        cancelledCourses: "Include",
-      }).then(normalizeResponse);
-      if (resp.schools.length) await doChunkUpsert(db, term, resp, null);
-      await sleep(1000);
+      const lower = sectionCodeBounds[i] as `${number}`;
+      const upper = (sectionCodeBounds[i + 1] ?? LAST_SECTION_CODE) as `${number}`;
+      await ingestChunk(db, term, lower, upper);
     }
   }
   await scrapeGEsForTerm(db, term);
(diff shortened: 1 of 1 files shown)


apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -830,16 +830,9 @@ export async function scrapeTerm(
...
-      const upper = sectionCodeBounds[i + 1] ?? LAST_SECTION_CODE;
-      const sectionCodes = `${lower}-${upper}`;
-      console.log(`Scraping chunk ${sectionCodes}`);
-      const resp = await request(term, {
-        sectionCodes,
-        cancelledCourses: "Include",
-      }).then(normalizeResponse);
-      if (resp.schools.length) await doChunkUpsert(db, term, resp, null);
-      await sleep(1000);
+      const lower = sectionCodeBounds[i] as `${number}`;
```
ecxyzzy, line 833: What is the purpose of this cast?

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -853,6 +846,48 @@ export async function scrapeTerm(
...
+     assuming network, etc. conditions are fine, we have more than 900 sections here
+     this means we somehow overran our 1% tolerance
+     that's okay; we can be suboptimal this time so we get all the sections that exist.
+     we're going to recompute the chunks at the start of the next scrape,
+     so that one will run optimally, given no such failure occurs again
+
+     we're going to bisect this chunk and try the two halves separately; eventually,
+     we'll have <= 900 valid sections in a chunk and we'll be in the clear
+    */
+    const lowerInt = baseTenIntOrNull(lower) as number;
```
ecxyzzy, line 875: If you're going to always cast to `number` I assume this means you expect `lower` and `upper` to always be a string containing a number. Therefore, you should use `Number.parseInt(x, 10)`.

Also not quite sure why `lower` and `upper` are being passed as `${number}` anyway, but that's addressed above.

ecxyzzy (changes requested):

apps/data-pipeline/websoc-scraper/src/lib.ts
laggycomputer: Wanted to assert this is a valid int.

ecxyzzy: As in, assert that it is a string containing a valid int?

…I guess, but this feels unnecessary. If it's because `ingestChunk` is exported and thus should have sensible inputs, it also doesn't need to be exposed.

laggycomputer: Without a type assertion, I cannot define the parameter types for `ingestChunk`. The use of raw SQL results in incorrect type inference.

If you want a less contrived typehint, `string` may be acceptable.

`ingestChunk` will not be exported.

ecxyzzy: Makes sense to me now - once `ingestChunk` is unexported then lgtm

ecxyzzy (approved):
lgtm

Issue #116, opened 2025-02-10
Need AP Exam name mappings


laggycomputer:
PeterPortal currently allows you to enter transfer and AP credits. These require that you type the UCI catalogue name of an AP Exam (their example is "AP COMP SCI A") instead of any sort of official name. We can remedy this:

* We can provide a source of truth of the official (College Board) names of these exams.
* By doing this we also enumerate every AP Exam which appears in the catalogue.

This is a source of truth of interest to API consumers and is thus under our mandate.

PR #117, opened 2025-02-10
feat: ap exam mappings

laggycomputer:
## Description

Provide mappings from UCI Catalogue names of AP Exams to official names as provided by College Board.

We add a new script to the extraction pipeline which simply dumps a hardcoded map into a new table, `ap_exam`. 
**This is a database schema change.**

## Related Issue

Execute #116.

## Motivation and Context

See issue.

## How Has This Been Tested?

There are three cases of interest:
* Passing no parameter, listing all mappings.
* Passing a valid AP Exam as appears in UCI Catalogue; this will retrieve the mapping for only this name.
* Passing an invalid name; this is a 404.

REST, no query:
Screenshot_20250210_122521

REST, valid name:
Screenshot_20250210_122535

REST, invalid name:
Screenshot_20250210_123341

GraphQL, no query:
Screenshot_20250210_122428

GraphQL, valid name:
Screenshot_20250210_122246

GraphQL, invalid name:
Screenshot_20250210_123504

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/base/response.ts
+++ b/apps/api/src/schema/base/response.ts
@@ -11,8 +11,11 @@ export const cursorResponseSchema = <T extends z.ZodType>(data: T) =>
     ok: z.literal<boolean>(true).openapi({}),
     data: z.object({
       items: data.openapi({ description: "The list of requested items" }),
-      nextCursor: z.string().nullable().openapi({
-        description: "Cursor pointing to the next page. Null if there are no more results",
-      }),
+      nextCursor: z
+        .string()
+        .nullable()
+        .openapi({
+          description: "Cursor pointing to the next page. Null if there are no more results",
+        }),
     }),
   });
--- a/apps/api/src/schema/base/response.ts
+++ b/apps/api/src/schema/base/response.ts
@@ -11,11 +11,8 @@ export const cursorResponseSchema = <T extends z.ZodType>(data: T) =>
     ok: z.literal<boolean>(true).openapi({}),
     data: z.object({
       items: data.openapi({ description: "The list of requested items" }),
-      nextCursor: z
-        .string()
-        .nullable()
-        .openapi({
-          description: "Cursor pointing to the next page. Null if there are no more results",
-        }),
+      nextCursor: z.string().nullable().openapi({
+        description: "Cursor pointing to the next page. Null if there are no more results",
+      }),
     }),
   });
--- a/apps/data-pipeline/ap-exams-importer/package.json
+++ b/apps/data-pipeline/ap-exams-importer/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "@apps/ap-exams-importer",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Script for importing AP Exam names from a statuc mapping",
+  "type": "module",
+  "scripts": {
+    "check:types": "tsc -p ./tsconfig.json -noEmit",
+    "start": "dotenv -e ../../../.env -- tsx src/index.ts"
+  },
+  "dependencies": {
+    "@packages/db": "workspace:*"
+  }
+}
--- a/apps/data-pipeline/ap-exams-importer/package.json
+++ b/apps/data-pipeline/ap-exams-importer/package.json
@@ -2,7 +2,7 @@
   "name": "@apps/ap-exams-importer",
   "version": "0.0.0",
   "private": true,
-  "description": "Script for importing AP Exam names from a statuc mapping",
+  "description": "Script for importing AP Exam names from a static mapping",
   "type": "module",
   "scripts": {
     "check:types": "tsc -p ./tsconfig.json -noEmit",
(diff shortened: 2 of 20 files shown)


apps/data-pipeline/ap-exams-importer/src/names.ts
```diff
@@ -0,0 +1,21 @@
+// https://apstudents.collegeboard.org/courses
+
+export default {
+  "AP CALCULUS AB": "AP Calculus AB",
+  "AP CALCULUS AB SUB": "AP Calculus BC, Calculus AB subscore",
```
ecxyzzy, line 5: I think this can and should be stated more concisely as "AP Calculus BC (AB subscore)". However, open to arguments for keeping it this way.

apps/data-pipeline/ap-exams-importer/src/names.ts
```diff
@@ -0,0 +1,21 @@
...
+  "AP CALCULUS BC": "AP Calculus BC",
+  "AP CHEMISTRY": "AP Chemistry",
+  "AP COMP SCI A": "AP Computer Science A",
+  // this hasn't been offered since 2009
+  "AP COMP SCI AB": "AP Computer Science AB",
```
ecxyzzy, line 10: Given that this hasn't been offered in 2009 and it looks similar to CSA, we should consider adding an indication to the human-readable name. Perhaps something like "AP Computer Science AB (last offered 2009)".

apps/data-pipeline/ap-exams-importer/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@apps/ap-exams-importer",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Script for importing AP Exam names from a statuc mapping",
```
ecxyzzy, line 5:
```suggestion
  "description": "Script for importing AP exam names from a static mapping",
```

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  await db
+    .insert(apExams)
+    .values(
+      Object.entries(apExamNames).map(([k, v]) => ({
+        id: k,
+        officialName: v,
+      })),
+    )
+    .onConflictDoUpdate({ target: apExams.id, set: conflictUpdateSetAllCols(apExams) });
+
```
ecxyzzy, line 24: A call to `db.$client.end` should be added here to free the connection used by the script.

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+async function main() {
+  const url = process.env.DB_URL;
+  if (!url) throw new Error("DB_URL not found");
+  const db = database(url);
+
+  await db
+    .insert(apExams)
+    .values(
+      Object.entries(apExamNames).map(([k, v]) => ({
```
ecxyzzy, line 18: When destructuring key-value tuples from `Object.entries`, the key and value can be named whatever you want. So you could theoretically call them `id` and `officialName`, then use the object initializer shorthand in the object returned by `Array.map`.

I don't know if this is actually any more concise though, I just like using the objinit shorthand.

apps/api/src/schema/base/response.ts
```diff
@@ -11,11 +11,8 @@ export const cursorResponseSchema = <T extends z.ZodType>(data: T) =>
...
       items: data.openapi({ description: "The list of requested items" }),
-      nextCursor: z
-        .string()
-        .nullable()
-        .openapi({
-          description: "Cursor pointing to the next page. Null if there are no more results",
-        }),
+      nextCursor: z.string().nullable().openapi({
+        description: "Cursor pointing to the next page. Null if there are no more results",
+      }),
```
ecxyzzy, line 16: This reformatting is out of scope for this PR. Not sure if deliberate or if Biome somehow didn't catch it before.

ecxyzzy (changes requested):
Minor nits abound as is my wont. Looks good otherwise though.

apps/data-pipeline/ap-exams-importer/package.json
laggycomputer: Typo, indeed, but "AP Exam" is the capitalization used by College Board.

apps/api/src/schema/base/response.ts
laggycomputer: Biome did this. It really is checked in misformatted right now.

apps/data-pipeline/ap-exams-importer/src/index.ts
laggycomputer: Don't really have an opinion; I'll do that.

apps/data-pipeline/ap-exams-importer/src/names.ts
laggycomputer: But then this wouldn't be the official name from College Board and we are now injecting opinion.

apps/data-pipeline/ap-exams-importer/src/names.ts
laggycomputer: The [College Board page](https://apstudents.collegeboard.org/about-ap-scores/special-score-structure-calculus-bc) calls it the
> Calculus AB subscore on the Calculus BC Exam

or the

> Calculus AB subscore

so I want to be as conservative as possible about how we put the AB and BC aspects together.

laggycomputer:
we're gonna give this a makeover

@laggycomputer ngmi

PR #122, opened 2025-02-15
feat(dw-scraper): better wildcard handling

### LINKED ISSUE #120: DegreeWorks wildcards not parsed correctly
In DegreeWorks, the "@" symbol sometimes appears in course specifications.
This indicates a wildcard.

We currently interpret this "@" as a wildcard for a \*single\* free character.
However, in cases such as the BioSci minor, it actually means any number of free characters.
We need to investigate what this "@" really means; does it correspond to `.`, `.+`, or `.*` in RegEx?

My current theory is `.+`.

**PeterPortal staging is returning incorrect information right now because of this.**

### LINKED ISSUE #123: [ETL/DegreeWorks] Dump all course numbers for department-wide wildcards

ecxyzzy:
## Description

This PR aims to address two key issues with our handling of wildcards.

### Department-wide wildcards

For reasons that were never documented, wildcards for an entire department with no other qualifier (e.g. BIOSCI @) were never properly handled. This PR implements handling for such cases by dumping all courses in the department. This is a substantial amount of data, but no feasible workaround exists as department-wide wildcards may cohabitate with other complex requirements.

While implementing this feature, I noted that the department code given by DegreeWorks is the shortened version (without whitespace). A generated column and index thereon is added so that this data is readily and correctly accessible by the scraper.

### Wildcard translation to RegExp

We also aim to correct our previous incorrect assumption of what the wildcard `@` translates to. After some investigation, we present and implement the following new assumption.

We define a **wildcard block** as one or more consecutive (i.e. not broken by whitespace) occurrences of `@`. A wildcard block that has _n_ occurrences of `@` translates to the regular expression `\w{n,}`. It is worth noting of course that `\w{1,}` collapses to `\w+`, which is similar to the theory posited in #120.

So far, we have only observed occurrences of wildcard blocks with _n_ `\in` [1, 2], so the implementation reflects this fact. We argue that the pathological cases of _n_ > 2 are unlikely to occur in practice as there are no four-digit course numbers, and upper division courses with significant prefix/suffix use in their numbers are almost always covered by the _n_ = 2 case.

## Related Issue

Closes #120; closes #123.

## Motivation and Context

fixes PP (allegedly)

## How Has This Been Tested?
Instrumented with logging locally. Degreeworks data in production is in the process of being refreshed.

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
+++ b/apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
@@ -55,12 +55,16 @@ export class AuditParser {
     // "ELECTIVE @" is typically used as a pseudo-course and can be safely ignored.
     if (courseIdLike.match(AuditParser.electiveMatcher)) return [];
     const [department, courseNumber] = courseIdLike.split(" ");
+    if (courseNumber === "@") {
+      // Department-wide wildcards.
+      return this.db.select().from(course).where(eq(course.shortenedDept, department));
+    }
     if (courseNumber.match(AuditParser.wildcardMatcher)) {
       // Wildcard course numbers.
       return await this.db
         .select()
         .from(course)
-        .where(eq(course.department, department))
+        .where(eq(course.shortenedDept, department))
         .then((rows) =>
           rows.filter((x) =>
             x.courseNumber.match(
@@ -76,7 +80,7 @@ export class AuditParser {
       return await this.db
         .select()
         .from(course)
-        .where(eq(course.department, department))
+        .where(eq(course.shortenedDept, department))
         .then((rows) =>
           rows.filter(
             (x) =>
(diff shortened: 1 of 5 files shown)


apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -64,9 +64,7 @@ export class AuditParser {
         .then((rows) =>
           rows.filter((x) =>
             x.courseNumber.match(
-              new RegExp(
-                `^${courseNumber.replace(/@+/g, `.{${[...courseNumber].filter((y) => y === "@").length},}`)}`,
-              ),
+              new RegExp(`^${courseNumber.replace("@@", "\\w{2,}").replace("@", "\\w+")}`),
```
laggycomputer, line 67:
```suggestion
              // we assume @@@ and longer never happen (#122)
              new RegExp(`^${courseNumber.replace("@@", "\\w{2,}").replace("@", "\\w+")}`),
```

laggycomputer (changes requested):

laggycomputer (approved):

PR #137, opened 2025-03-05
feat: added base scraper for map-scraper

waterkimchi:
## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/map-scraper/src/index.ts
+++ b/apps/data-pipeline/map-scraper/src/index.ts
@@ -1,0 +1,139 @@
+import { exit } from "node:process";
+import { doScrape } from "$lib";
+import { database } from "@packages/db";
+import winston from "winston";
+
+interface LocationListData {
+  catId: number;
+  lat: number;
+  lng: number;
+  id: number;
+  name: string;
+}
+
+interface LocationDetailData {
+  mediaUrlTypes: string[];
+  mediaUrls: string[];
+}
+
+interface BuildingLocation {
+  name: string;
+  lat: number;
+  lng: number;
+  imageURLs: string[];
+}
+
+const defaultFormat = [
+  winston.format.timestamp(),
+  winston.format.printf((info) => `[${info.timestamp} ${info.level}] ${info.message}`),
+];
+
+const logger = winston.createLogger({
+  level: "info",
+  format: winston.format.combine(...defaultFormat),
+  transports: [
+    new winston.transports.Console({
+      format: winston.format.combine(winston.format.colorize(), ...defaultFormat),
+    }),
+    new winston.transports.File({
+      filename: `${__dirname}/../logs/${Date.now()}.log`,
+    }),
+  ],
+});
+
+const CATEGORIES: Set<number> = new Set([
+  8424, 8309, 8311, 8392, 8405, 44392, 44393, 44394, 44395, 44396, 44397, 44398, 44400, 44401,
+  44402, 44538, 44537, 44399, 8396, 11907, 8400, 10486, 11906, 11889, 8310, 8312, 8393, 8394, 8397,
+  8398, 8399, 8404, 8407, 8408, 11891, 11892, 11899, 11900, 11902, 21318, 8406, 11908, 11935,
+]);
+const LOCATIONS_LIST_API =
+  "https://api.concept3d.com/locations?map=463&key=0001085cc708b9cef47080f064612ca5";
+const LOCATIONS_DETAIL_API = (id: number) =>
+  `https://api.concept3d.com/locations/${id}?map=463&key=0001085cc708b9cef47080f064612ca5`;
+const locationsCatalogue: Record<number, BuildingLocation> = {};
+const locationIds: Record<string, number> = {};
+
+function sleep(ms: number): Promise<number> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function fetchLocations() {
+  const locationsListResponse: Response = await fetch(LOCATIONS_LIST_API);
+  const locations: LocationListData[] = (await locationsListResponse.json()) as LocationListData[];
+  console.log(locations);
+  // Go through all the locations
+  for (const location of locations) {
+    // Check if location was already recorded
+    const locationExists: boolean = Object.values(locationsCatalogue).some(
+      (existingLocation) => existingLocation.name === location.name,
+    );
+
+    if (!CATEGORIES.has(location.catId)) return;
+
+    if (locationExists) return;
+
+    // If one of the locations belongs to one of the categories above, and is not a duplicate
+    // Hits location data API to get details on location
+    const locationsDetailResponse: Response = await fetch(LOCATIONS_DETAIL_API(location.id));
+    const locationData: LocationDetailData =
+      (await locationsDetailResponse.json()) as LocationDetailData;
+
+    const imgUrls: string[] = [];
+
+    // Collects image URLs by checking if type is an image
+    // { 'mediaUrlTypes': ['image', '...', 'image'], 'mediaUrls': ['xyz.jpg', '...', 'abc.png']}
+    if (locationData.mediaUrlTypes !== undefined) {
+      for (const [i, media] of locationData.mediaUrlTypes.entries()) {
+        if (media === "image") {
+          imgUrls.push(locationData.mediaUrls[i]);
+        }
+      }
+    }
+
+    locationsCatalogue[location.id] = {
+      name: location.name,
+      lat: location.lat,
+      lng: location.lng,
+      imageURLs: imgUrls,
+    };
+
+    const locationName: string = location.name.includes("(")
+      ? location.name.substring(location.name.indexOf("(") + 1, location.name.indexOf(")"))
+      : location.name;
+    locationIds[locationName] = location.id;
+
+    await sleep(250);
+  }
+}
+
+async function main() {
+  const url = process.env.DB_URL;
+  if (!url) throw new Error("DB_URL not found");
+  const db = database(url);
+
+  logger.info("map-scrapper starting...");
+  await fetchLocations();
+  const buildingInterfaceStr = `
+export interface Building {
+imageURLs: string[];
+lat: number;
+lng: number;
+name: string;
+}\n\n`;
+
+  const buildingCatalogueStr = `
+const buildingCatalogue: Record<string, Building> = ${JSON.stringify(
+    locationsCatalogue,
+    null,
+    4,
+  )};\nexport default buildingCatalogue;`;
+
+  const locationStr = `
+const locations: Record<string, number> = ${JSON.stringify(locationIds, null, 4)};
+\nexport default locations;`;
+  logger.info(buildingInterfaceStr + buildingCatalogueStr + locationStr);
+  await doScrape(db);
+  exit(0);
+}
+
+main().then();
--- a/apps/data-pipeline/map-scraper/package.json
+++ b/apps/data-pipeline/map-scraper/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "@apps/map-scraper",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Scraper for campus map data",
+  "type": "module",
+  "scripts": {
+    "check:types": "tsc -p ./tsconfig.json -noEmit",
+    "deploy": "wrangler deploy",
+    "postinstall": "wrangler types --x-include-runtime",
+    "start": "dotenv -e ../../../.env -- tsx src/index.ts"
+  },
+  "dependencies": {
+    "@packages/db": "workspace:*",
+    "@packages/stdlib": "workspace:*",
+    "winston": "3.15.0"
+  },
+  "devDependencies": {
+    "wrangler": "3.100.0"
+  }
+}
--- a/apps/data-pipeline/map-scraper/tsconfig.json
+++ b/apps/data-pipeline/map-scraper/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "../../../tsconfig.json"
+}


laggycomputer:
can you fix the merge conflict?

waterkimchi:
> can you fix the merge conflict?

completed.

apps/api/src/rest/routes/map.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const mapRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
+
+const mapLocationsRoute = createRoute({
+  summary: "Retrieve coordinates for locations",
+  operationId: "getLocations",
+  tags: ["Map"],
+  method: "get",
+  path: "/",
+  description: "List all available map data from UCI's interactive map'",
```
laggycomputer, line 16: spare quote here

apps/api/src/rest/routes/map.ts
```diff
@@ -0,0 +1,48 @@
...
+import { productionCache } from "$middleware";
+import { errorSchema, mapQuerySchema, mapResponseSchema, responseSchema } from "$schema";
+import { MapService } from "$services";
+import { OpenAPIHono, createRoute } from "@hono/zod-openapi";
+import { database } from "@packages/db";
+
+const mapRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
+
+const mapLocationsRoute = createRoute({
+  summary: "Retrieve coordinates for locations",
```
laggycomputer, line 11: this is a bit vague; can we specify e.g. UCI locations?

apps/api/src/schema/map.ts
```diff
@@ -0,0 +1,29 @@
...
+  z.object({
+    id: z.string().openapi({
+      description: "ID of this location",
+      example: "83021",
+    }),
+    name: z.string().openapi({
+      description: "The generic name of the location",
+      example: "Art Studio (ART)",
+    }),
+    latitude: z.string().openapi({
```
laggycomputer, line 20: these should be numbers

apps/api/src/schema/map.ts
```diff
@@ -0,0 +1,29 @@
...
+      description: "The latitude of the selected location",
+      example: "33.650162",
```
laggycomputer, line 22: and if you re-type these number, the example should be a number too

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+interface BuildingLocation {
+  name: string;
+  lat: number;
+  lng: number;
+  imageURLs: string[];
+}
+
+const CATEGORIES: Set<number> = new Set([
```
laggycomputer, line 25: can you leave a note about how we decided that hard-coding these is mandatory?

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+          target: mapLocation.id,
+          set: {
+            name: locationData.name,
+            latitude: locationData.lat.toString(),
+            longitude: locationData.lng.toString(),
+          },
+        });
+    }
+  });
+  const buildingInterfaceStr = `
```
laggycomputer, line 118: why did we need to print this out?

laggycomputer (changes requested):
nits for now, will proceed to further testing after one round when i understand the code a bit better

apps/data-pipeline/map-scraper/src/index.ts
waterkimchi: It was for debugging purposes, but removed from recent commit.

apps/api/src/services/map.ts
```diff
@@ -0,0 +1,26 @@
...
+        name: mapLocation.name,
+        latitude: mapLocation.latitude,
+        longitude: mapLocation.longitude,
+      })
+      .from(mapLocation)
+      .where(query.id ? eq(mapLocation.id, query.id) : undefined);
+
+    return results.map((r) => ({
+      ...r,
+      latitude: Number(r.latitude),
```
laggycomputer, line 22: why doesn't this mapping happen automatically?

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const CATEGORIES: Set<number> = new Set([
+  8424, 8309, 8311, 8392, 8405, 44392, 44393, 44394, 44395, 44396, 44397, 44398, 44400, 44401,
+  44402, 44538, 44537, 44399, 8396, 11907, 8400, 10486, 11906, 11889, 8310, 8312, 8393, 8394, 8397,
+  8398, 8399, 8404, 8407, 8408, 11891, 11892, 11899, 11900, 11902, 21318, 8406, 11908, 11935,
+]);
+
+// The Concept3D API key is hardcoded here, which is generally a bad practice due to the security risks.
+// For truly sensitive keys, environment variables or a secrets manager would be used (as with DB_URL).
+// This approach was considered pragmatic for this specific utility and non-sensitive key.
```
laggycomputer, line 33: comment doesn't need to this verbose, just indicate you had no choice besides hardcoding and you will investigate another solution eventually

the idea that hardcoding keys is bad is common knowledge so you don't have to say it here

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const LOCATIONS_LIST_API =
+  "https://api.concept3d.com/locations?map=463&key=0001085cc708b9cef47080f064612ca5";
+const LOCATIONS_DETAIL_API = (id: number) =>
+  `https://api.concept3d.com/locations/${id}?map=463&key=0001085cc708b9cef47080f064612ca5`;
+const locationsCatalogue: Record<number, BuildingLocation> = {};
+const locationIds: Record<string, number> = {};
+
+function sleep(ms: number): Promise<number> {
```
laggycomputer, line 42: why can't this be reused from elsewhere?

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  8398, 8399, 8404, 8407, 8408, 11891, 11892, 11899, 11900, 11902, 21318, 8406, 11908, 11935,
+]);
+
+// The Concept3D API key is hardcoded here, which is generally a bad practice due to the security risks.
+// For truly sensitive keys, environment variables or a secrets manager would be used (as with DB_URL).
+// This approach was considered pragmatic for this specific utility and non-sensitive key.
+
+const LOCATIONS_LIST_API =
+  "https://api.concept3d.com/locations?map=463&key=0001085cc708b9cef47080f064612ca5";
+const LOCATIONS_DETAIL_API = (id: number) =>
```
laggycomputer, line 37: this is misleadingly named like a constant despite being a function; either name it in camelCase such that it is clearly a function or make it a true string constant

apps/data-pipeline/map-scraper/src/index.ts
laggycomputer: unresolving - you need to explicitly make it clear that you had to hardcode the api key AND non-secret constants like the categories and UCI map ID

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+function sleep(ms: number): Promise<number> {
+  return new Promise((resolve) => setTimeout(() => resolve(ms), ms));
+}
+
+async function fetchLocations() {
+  const locationsListResponse: Response = await fetch(LOCATIONS_LIST_API);
+  const locations: LocationListData[] = (await locationsListResponse.json()) as LocationListData[];
+  for (const location of locations) {
+    // Check if location was already recorded
+    const locationExists: boolean = Object.values(locationsCatalogue).some(
```
laggycomputer, line 51: vaguely named - all locations "exist", but what you're checking here is that the location has *already been acquired*

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      continue;
+    }
+    if (locationExists) {
+      continue;
+    }
+
+    // If one of the locations belongs to one of the categories above, and is not a duplicate
+    // Hits location data API to get details on location
+    const locationsDetailResponse: Response = await fetch(LOCATIONS_DETAIL_API(location.id));
+    const locationData: LocationDetailData =
```
laggycomputer, line 65: it should not be necessary to both type the variable and cast the expression you assign into it

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      (await locationsDetailResponse.json()) as LocationDetailData;
+
+    const imgUrls: string[] = [];
+
+    // Collects image URLs by checking if type is an image
+    // { 'mediaUrlTypes': ['image', '...', 'image'], 'mediaUrls': ['xyz.jpg', '...', 'abc.png']}
```
laggycomputer, line 71: the array structure here makes me a bit uncomfortable; can you store a list of (type, url) tuples?

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    locationsCatalogue[location.id] = {
+      name: location.name,
+      lat: location.lat,
+      lng: location.lng,
+      imageURLs: imgUrls,
+    };
+
+    const locationName: string = location.name.includes("(")
```
laggycomputer, line 87: this is probably more concise with `.split` and shouldn't need to be explicitly typed either

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+async function main() {
+  const url = process.env.DB_URL;
+  if (!url) throw new Error("DB_URL not found");
+  const db = database(url);
+  console.log("map scraper starting...");
+  await fetchLocations();
+  console.log("Done fetching locations...");
+  await db.transaction(async (tx) => {
+    for (const locationId in locationsCatalogue) {
+      const locationData = locationsCatalogue[locationId];
```
laggycomputer, line 103: why doesn't `Object.entries` fit your use case here?

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      await tx
+        .insert(mapLocation)
+        .values({
+          id: locationId,
+          name: locationData.name,
+          latitude: locationData.lat.toString(),
```
laggycomputer, line 109: is `toString()` still sound if the type is now `Number`?

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+          longitude: locationData.lng.toString(),
+          imageURLs: locationData.imageURLs,
+        })
+        .onConflictDoUpdate({
```
laggycomputer, line 113: why not update every column on conflict? is there a good reason to retain the `imageURLs` column on conflict?

packages/db/migrations/0017_map_locations.sql
laggycomputer, line 1: the next snapshot is 16; we can't assume #169 will merge first and this will merge second
when one merges, redo the drizzle stuff on the other PR
until then, both should be ready to merge immediately

packages/db/src/schema.ts
```diff
@@ -707,6 +708,15 @@ export const studyRoomSlot = pgTable(
   ],
 );
 
+// Map data
+export const mapLocation = pgTable("map_location", {
+  id: numeric("id").primaryKey(),
+  name: varchar("name").notNull(),
+  latitude: numeric("latitude").notNull(),
+  longitude: numeric("longitude").notNull(),
+  imageURLs: json("image_urls").$type<string[]>(),
```
laggycomputer, line 717: why can't this be a SQL array of `varchar`?

laggycomputer (changes requested):
looked into scraper side more deeply this time

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+const seenIds = new Set<string>();
+
+async function fetchImageUrls(id: string): Promise<string[]> {
+  try {
+    const res = await fetch(getLocationDetail(id));
+    if (!res.ok) {
+      console.warn(`[detail] ${id} HTTP ${res.status}`);
+      return [];
+    }
+    const data = (await res.json()) as LocationDetailData;
```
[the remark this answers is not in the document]
sanskarm7: ive tried to implement zod, however i haven't done so before and am unsure if i followed convention

apps/data-pipeline/map-scraper/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  const db = database(url);
+
+  console.log("started map scraper");
+  await fetchLocations();
+  console.log("upserting…");
+
+  let upserts = 0;
+
+  await db.transaction(async (tx) => {
+    for (const [id, data] of locationsCatalogue) {
```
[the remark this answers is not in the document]
sanskarm7: i've leaned on the side continuing to use a transaction for safety in the event of failure. it now creates a single list, deletes conflicting entries, then does one upsert.

sanskarm7:
This PR is being closed for now due to the increased operational overhead of introducing PostGIS into our RDS instance. The original plan for enhancing the Concept3D API was to use PostGIS to efficiently store and query geospatial data (e.g., k-nearest-neighbor queries, point-in-polygon checks, and similar coordinate-based operations).

However, the intended consumer of this endpoint, ZotNFound, no longer requires this data. As a result, this work is being paused until geospatial features are needed again.

For future maintainers:

 - You will need to investigate the process of enabling the PostGIS extension on RDS and how to incorporate it into our migration workflow.

 - Concept3D’s API also required several hard-coded parsing rules because their distinction between “visible” labels and functional labels was not clearly documented.

This context should help anyone who revisits and continues this endpoint in the future.

Issue #106, opened 2025-02-02
Create scraper and endpoint for library busyness data


andrew-wang0:
Create a library busyness endpoint that provides current and historical (hourly?) data on library floor busyness
Our best best is likely to use the api at `https://waitz.io/live/irvine`

ecxyzzy:
https://www.lib.uci.edu/where-do-you-want-study-today

PR #149, opened 2025-04-23
feat: web scraper and endpoint for library traffic data

estao1:
## Description

Adds a new library-traffic-scraper service to automatically collect and update library traffic data from the [UCI Libraries library traffic site](https://www.lib.uci.edu/where-do-you-want-study-today). This includes scheduled scraping into a latest library traffic and an archived library traffic database table, and an endpoint to retrieve the latest traffic data with a location parameter.

## Related Issue

https://github.com/icssc/anteater-api/issues/106

## Motivation and Context

This scraper provides a backend foundation for displaying current and historical library floor traffic data. It's meant to support analytics, dashboards, and future integrations across other systems using the API.

## How Has This Been Tested?

- Connected to local Postgres DB (anteater_api) and verified table schema + data.
- Manually ran `npm start` to scrape and add live + historical data.
- Ensured endpoint on API docs returned accurate information.
- Ran `npx wrangler dev --env development` and confirmed successful scheduled run.
- Queried data via GraphQL and confirmed accurate filtering.

image

image

image

## Types of changes

## Checklist:


apps/api/src/schema/library-traffic.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from "@hono/zod-openapi";
+
+export const libraryTrafficQuerySchema = z.object({
+  locationName: z
+    .string()
+    .openapi({
+      param: { name: "locationName", in: "query" },
```
laggycomputer, line 7: this `param` is redundant and we should have an example and description

apps/api/src/schema/library-traffic.ts
```diff
@@ -0,0 +1,20 @@
...
+    })
+    .optional(),
+});
+
+export const libraryTrafficSchema = z.object({
```
laggycomputer, line 12: we typically have descriptions on these fields; i argue the meaning here isn't 
perfectly obvious from just names and examples (even if i have a 75% idea) so you should add descriptions here

apps/api/src/services/library-traffic.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { database } from "@packages/db";
+import { eq } from "@packages/db/drizzle";
+import { libraryTraffic } from "@packages/db/schema";
+import type { z } from "zod";
+
+type LibraryTrafficServiceInput = z.infer<typeof libraryTrafficQuerySchema>;
+
+type LibraryTrafficServiceOutput = z.infer<typeof libraryTrafficSchema>;
+
+const transformTrafficRow = (
```
laggycomputer, line 11: why isn't this covered by either zod schema parsing or processing you can do in the SQL query?

apps/api/src/services/library-traffic.ts
```diff
@@ -0,0 +1,34 @@
...
+});
+
+export class LibraryTrafficService {
+  constructor(private readonly db: ReturnType<typeof database>) {}
+
+  async getLibraryTraffic(
+    input: LibraryTrafficServiceInput,
+  ): Promise<LibraryTrafficServiceOutput[]> {
+    const rows = await this.db
+      .select()
```
laggycomputer, line 28: you should prefer including `libraryTraffic.$inferSelect` for readability

apps/data-pipeline/library-traffic-scraper/src/handler.ts
```diff
@@ -0,0 +1,15 @@
...
+
+interface Env {
+  DB_URL: string;
+}
+
+export default {
+  async scheduled(_: ScheduledController, env: Env): Promise<void> {
+    await doScrape(env.DB_URL);
+  },
+  async fetch(_: Request, env: Env): Promise<Response> {
```
laggycomputer, line 11: do you intend to fetch to this worker? if so, why?

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,136 @@
+import { database } from "@packages/db";
+import { libraryTraffic, libraryTrafficHistory } from "@packages/db/schema";
+import fetch from "cross-fetch";
+
+const locationIds = [
```
laggycomputer, line 5: this is fragile; is there no other way to determine these IDs?

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,136 @@
...
+  "672",
+  "666",
+  "673",
+  "669",
+  "786",
+  "674",
+  "787",
+];
+
+const floors = [
```
laggycomputer, line 30: same as with `locationIds`

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,136 @@
...
+  };
+};
+
+type RawRespErr = { error: string };
+
+async function fetchLocation(id: string): Promise<RawRespOK["data"] | null> {
+  const url = `https://www.lib.uci.edu/sites/all/scripts/occuspace.php?id=${id}`;
+  try {
+    const txt = await fetch(url).then((r) => r.text());
+    let parsed = JSON.parse(txt);
```
laggycomputer, line 74: why not directly `await fetch(url).then(r => r.json())`? is there any special handling to be done? imo at most, split the fetch from parsing json and handle those separately

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    if (typeof parsed === "string") {
+      parsed = JSON.parse(parsed);
```
laggycomputer, line 77: and if we stay with this solution, i'm very uneasy about the use of `let` and the conditional alteration of this variable's type

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export async function doScrape(DB_URL: string) {
+  const db = database(DB_URL);
+  console.log("Starting library traffic scrape.");
+
+  for (const { id } of locationMappings) {
+    const data = await fetchLocation(id);
+    if (!data) {
+      console.error(`Failed to fetch data for location ID ${id}`);
```
laggycomputer, line 98: you catch and don't rethrow errors inside `fetchLocation` so they never propagate here

if there is a different unexpected state that really would send an exception out here, can you be more specific about it?

packages/db/src/schema.ts
```diff
@@ -743,6 +743,32 @@ export const apExamReward = pgTable("ap_exam_reward", {
...
+export const libraryTraffic = pgTable(
+  "library_traffic",
+  {
+    id: integer("id").primaryKey(),
+    locationName: varchar("location_name").notNull(),
+    trafficCount: integer("traffic_count").notNull(),
+    trafficPercentage: decimal("traffic_percentage").notNull(),
+    timestamp: timestamp("timestamp").notNull(),
+  },
+  (table) => [index().on(table.id), index().on(table.locationName)],
```
laggycomputer, line 755: you should not explicitly redeclare an index on a primary key (this is always created for you)

laggycomputer (changes requested):
some initial comments

apps/data-pipeline/library-traffic-scraper/src/lib.ts
estao1: I initially tried that, but the response from the `occuspace.php` endpoint is a double-encoded JSON, so calling `r.json()` doesn't give me the actual JSON object until I parse it twice. I can try cleaning up the fetch/parsing flow to make it clearer if needed.

apps/api/src/schema/library-traffic.ts
```diff
@@ -3,18 +3,39 @@ import { z } from "@hono/zod-openapi";
 export const libraryTrafficQuerySchema = z.object({
   locationName: z
     .string()
+    .describe("Filter results by exact location name")
```
laggycomputer: convention is to use `description` property of param to `openapi` (as you do below)

apps/api/src/schema/library-traffic.ts
```diff
@@ -3,18 +3,39 @@ import { z } from "@hono/zod-openapi";
...
+  trafficPercentage: z
+    .union([z.string(), z.number().min(0).max(1)])
+    .transform((v) => Number(v))
+    .openapi({
+      example: 0.33,
+      description: "Occupancy as a decimal percentage of total capacity (0 to 1)",
+    }),
+
+  timestamp: z
+    .union([z.string(), z.date()])
```
laggycomputer, line 35: not sure why we'd have this union - neither on REST nor GraphQL will the user ever be able to supply a constructed `Date` object

iirc we use `z.coerce.date()` elsewhere; see if that works for you

apps/api/src/services/library-traffic.ts
```diff
@@ -8,27 +8,23 @@ type LibraryTrafficServiceInput = z.infer<typeof libraryTrafficQuerySchema>;
...
+        locationName: libraryTraffic.locationName,
+        trafficCount: libraryTraffic.trafficCount,
+        trafficPercentage: libraryTraffic.trafficPercentage,
+        timestamp: libraryTraffic.timestamp,
+      })
       .from(libraryTraffic)
       .where(input.locationName ? eq(libraryTraffic.locationName, input.locationName) : undefined);
 
-    return rows.map(transformTrafficRow);
+    return rows.map((row) => libraryTrafficSchema.parse(row));
```
laggycomputer, line 28: two things:
1. by convention, the REST handler and GraphQL resolver should call `.parse` on some schema to validate what they get back from the service
2. you should redefine `libraryTrafficSchema` as an array of some other object (say, `libraryTrafficEntry`), and parse that single schema against the returned array of row objects

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export async function doScrape(DB_URL: string): Promise<void> {
+  console.log("Starting library traffic scrape.");
+
+  const pageHtml = await fetch("https://www.lib.uci.edu/where-do-you-want-study-today").then((r) =>
+    r.text(),
+  );
+
+  const scriptBlockMatch = pageHtml.match(
+    /<script\b[^>]*>([\s\S]*?locationIds[\s\S]*?)<\/script>/i,
```
laggycomputer, line 46: is there a good reason to not use `cheerio` here? it would probably make your life drastically easier and make this code more readable

laggycomputer (changes requested):
cooking

apps/data-pipeline/library-traffic-scraper/src/lib.ts
laggycomputer: holy smokes, that's scuffed

i'd still argue for considering both layers of parsing as one unit that can either entirely succeed or fail; if it fails in the middle there's probably an issue with the scraping logic that can be investigated further

but for code we place down for the long-term, i'd prefer one unit

estao1: Not sure if I understand, how can I change my implementation to be one unit?

laggycomputer: as in, one unit which unwraps both JSON layers and either succeeds entirely or fails entirely

apps/api/src/services/library-traffic.ts
```diff
@@ -0,0 +1,27 @@
...
+        id: libraryTraffic.id,
+        locationName: libraryTraffic.locationName,
+        trafficCount: libraryTraffic.trafficCount,
+        trafficPercentage: libraryTraffic.trafficPercentage,
+        timestamp: libraryTraffic.timestamp,
+      })
+      .from(libraryTraffic)
+      .where(input.locationName ? eq(libraryTraffic.locationName, input.locationName) : undefined);
+
+    return libraryTrafficSchema.parse(rows);
```
laggycomputer, line 25: the resolvers (REST and GraphQL) typically do this schema parsing step themselves; i agree that pattern has downsides but we should stick with it for now

packages/db/migrations/meta/0012_snapshot.json
laggycomputer, line 1: not sure how this patch happened (biome?), and it shouldn't be included in principle, but i personally don't really care

apps/data-pipeline/library-traffic-scraper/wrangler.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[observability]
+enabled = true
+
+[[hyperdrive]]
+binding = "DB"
+id = "1c4c82477bfa4114a48c63d4dc838e73"
+
+[env.development.vars]
+DB_URL = "postgres://postgres:postgres@localhost:5432/anteater_api"
```
laggycomputer, line 19: we never set this var in any other package in the import layer, and it's not good practice to make assumptions about people's dev configurations which are intentionally allowed to vary widely

laggycomputer (changes requested):

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+  return null;
+}
+
+function extractLocationIds(pageHtml: string): string[] {
+  const $ = load(pageHtml);
+
+  let locationIds: string[] = [];
+
+  $("script").each((_, scriptTag) => {
```
laggycomputer, line 44: the use of `let` can be avoided; can you use `find` to get a matching `<script>` and then set `const locationIds` using that single found script? you can still throw if you didn't find that `<script>` that you wanted

apps/api/src/schema/library-traffic.ts
```diff
@@ -0,0 +1,43 @@
...
+    description: "Name of the library location",
+  }),
+
+  trafficCount: z.number().int().nonnegative().openapi({
+    example: 57,
+    description: "Number of people currently detected at the location",
+  }),
+
+  trafficPercentage: z
+    .union([z.string(), z.number().min(0).max(1)])
```
laggycomputer, line 27: this should be written `z.coerce.number()`

in fact, since this schema will handle your outgoing result, there should be no need to coerce at all since you are exactly certain of the type

note also that `.min` and `.max` will be enforced but are not visible on the openapi schema at all

apps/api/src/services/library-traffic.ts
```diff
@@ -0,0 +1,27 @@
...
+type LibraryTrafficServiceInput = z.infer<typeof libraryTrafficQuerySchema>;
+type LibraryTrafficServiceOutput = typeof libraryTraffic.$inferSelect;
+
+export class LibraryTrafficService {
+  constructor(private readonly db: ReturnType<typeof database>) {}
+
+  async getLibraryTraffic(
+    input: LibraryTrafficServiceInput,
+  ): Promise<LibraryTrafficServiceOutput[]> {
+    return await this.db
```
laggycomputer, line 16: no need to await here, and if you make that change the function doesn't need to be async either

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,112 @@
...
+    console.log(`Fetched "${data.name}": count=${data.count}, pct=${data.percentage}%`);
+
+    await db
+      .insert(libraryTraffic)
+      .values([
+        {
+          id: data.id,
+          locationName: data.name,
+          trafficCount: data.count,
+          trafficPercentage: String(data.percentage),
```
laggycomputer, line 88: holup why is this being turned into a string when it's a float in postgres

laggycomputer (changes requested):
we're very close

laggycomputer (approved):
ship it

sanskarm7:
The code implementation was designed well and accomplished accurate scraping of the library busyness. However, the way it was formatted left a lot of ambiguity as to what information the endpoint was returning.

As a result, I have rewritten the scraper and refactored some of the endpoint. The overarching changes are as follows-

- The scraper now records "libraryName", which is inserted into each entry into the library traffic table
- The scraper now records "sub text" in associated rooms, and concatenates them in a locationName, so there are no more duplicates (For example: "4th Floor - Loft Study Zone")
- The libraryTrafficHistory is unchanged largely - we maintain storing library information by IDs

Action Items:
- GraphQL may still be broken
- We need to change libraryName to be required, and to give appropriate error messages if there are issues
- Extensive testing during different times of day, some room numbers like "678" in Science Library error out - need to look into why

apps/api/src/rest/routes/library-traffic.ts
```diff
@@ -0,0 +1,55 @@
...
+
+libraryTrafficRouter.openapi(libraryTrafficRoute, async (c) => {
+  const query = c.req.valid("query");
+  const service = new LibraryTrafficService(database(c.env.DB.connectionString));
+  const res = await service.getLibraryTraffic(query);
+
+  if (res.length === 0) {
+    return c.json(
+      { ok: false, message: "Library traffic data not found: check for typos in query" },
+      422,
```
laggycomputer, line 48: you throw 422 here, but a 404 equivalent on graphql; you should decide on one and stick with it

apps/api/src/schema/library-traffic.ts
```diff
@@ -0,0 +1,50 @@
...
+    example: "4th Floor - Nordstrom Honors Study Room",
+    description: "Name of the floor / section inside the library",
+  }),
+  trafficCount: z.number().int().nonnegative().openapi({
+    example: 57,
+    description: "Number of people currently detected at the location",
+  }),
+  trafficPercentage: z.coerce
+    .number()
+    .transform((v) => Number(v))
```
laggycomputer, line 36: transformation is redundant if coercing

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,196 @@
...
+    if (parsed.data && typeof parsed.data === "object") return (parsed as RawRespOK).data;
+    console.warn(`ID ${id} responded with error: ${(parsed as RawRespErr).error}`);
+  } catch (err) {
+    console.error(`Unexpected error while fetching ID ${id}:`, err);
+    throw err;
+  }
+  return null;
+}
+
+type ScrapeStatus = "active" | "idle" | "skip";
```
laggycomputer, line 100: can you explain at a high level in comments what these statuses mean?

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,196 @@
...
+  console.log("Starting library traffic scrape.");
+  const db = database(DB_URL);
+  const lookup = await collectLocationMeta();
+
+  const now = new Date();
+  const minutes = now.getMinutes();
+
+  for (const [id, meta] of Object.entries(lookup)) {
+    const status = getScrapeStatus(
+      meta.libraryName === "Langson Library"
```
laggycomputer, line 137: you should use an object literal as a lookup table here

laggycomputer (changes requested):

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,196 @@
...
+          ? "SL"
+          : "LGSC",
+    );
+
+    if (status === "skip") {
+      console.error(`Skipping ${meta.libraryName} — unknown library code.`);
+      continue;
+    }
+
+    if (status === "idle" && minutes !== 0) {
```
laggycomputer, line 149: you have no guarantee that this exact line will be trigger on the minute; your job could for example be started at 00:00:59 and reach here within the first minute of the hour, meaning this check would not do what you expected

laggycomputer (changes requested):

apps/data-pipeline/library-traffic-scraper/src/lib.ts
estao1: I added a 10 minute leeway period as a lightweight workaround, but lmk if you'd want something more robust and if so, if you have any ideas on how I could implement it.

laggycomputer: isn't the scraping interval 15? i'd prefer we scrape slightly too often rather than miss an hour a few times

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -129,24 +132,24 @@ export async function doScrape(DB_URL: string) {
...
 
   for (const [id, meta] of Object.entries(lookup)) {
-    const status = getScrapeStatus(
-      meta.libraryName === "Langson Library"
-        ? "LL"
-        : meta.libraryName === "Science Library"
-          ? "SL"
-          : "LGSC",
-    );
+    const libraryCodeMap: Record<string, "LL" | "SL" | "LGSC"> = {
```
laggycomputer, line 139: you can simplify this type hint by casting the object literal to `const`, giving you maximal type information

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -129,24 +132,24 @@ export async function doScrape(DB_URL: string) {
...
+
+    const status = getScrapeStatus(libraryCodeMap[meta.libraryName]);
 
     if (status === "skip") {
       console.error(`Skipping ${meta.libraryName} — unknown library code.`);
       continue;
     }
 
-    if (status === "idle" && minutes !== 0) {
+    if (status === "idle" && minutes > 5 && minutes < 55) {
```
laggycomputer, line 152: i think this should be the interval [0, 15) minutes and it should be commented exactly why we need that, for posterity

laggycomputer (changes requested):

apps/data-pipeline/library-traffic-scraper/src/lib.ts
estao1: I set it to [0, 15), wouldn't [0,16) enable 2 scrapes per hour instead of the expected 1?

laggycomputer: yeah that's fine

laggycomputer (approved):

Issue #193, opened 2025-08-24
course scraper not configured for new version of uci catalogue site


sanskarm7:
The existing implementation for the course scraper relies on new line splits between sections of information. UCI has since changed their HTML which is making the scraper fail

Issue #192, opened 2025-08-19
Prerequisites for ENGRMAE 151A/B are incorrect


Awesome-E:
Expected Data: ENGRMAE 120 is a co-requisite course for ENGRMAE 151B (see [ENGRMAE Catalogue](https://catalogue.uci.edu/allcourses/engrmae/#:~:text=Mechanical%20Engineering%20Design))

Actual Data: ENGRMAE 120 is instead listed as a co-requisite course for ENGRMAE 151A

This may simply require updating the database by re-running the scraper.

Awesome-E:
ENGRMAE 151B is still not listing ENGRMAE 120A as a corequisite in the prerequisite tree, although it is in the `corequisite` string

laggycomputer:
WebSoc and the Catalogue currently disagree; WebSoc does not acknowledge this co-req. Need to Monitor the Situation further

Issue #152, opened 2025-04-30
[ETL/Courses] Ensure department aliases are not overwritten


ecxyzzy:
We currently define the following department aliases for the search endpoint to use:

* CS - COMPSCI
* ESS - EARTHSS
* ICS - I&C SCI
* INF - IN4MATX
* MAE - ENGRMAE
* WR - WRITING

When running the course scraper, courses that have updated info have their department aliases erased. This is not desirable behavior and can be mitigated by modifying the scraper.

PR #194, opened 2025-08-24
fix: course scraper adjustment for "courseblock"  change

sanskarm7:
## Description

rewrote `scrapeCoursesInDepartment` to parse per-course `.courseblock` elements and their `.detail-*` children instead of relying on newline-splitting. This matches the current catalogue HTML and removes brittleness around formatting changes.

also addressed missing `departmentAlias` in course by adding a mapping on completed scrape

## Related Issue

addresses #192, #193, #195, #152

## Motivation and Context

course scraper needs to be fixed to run for the 25/26 academic year

## How Has This Been Tested?

current testing status shows that entries for courses are accurately updated with the exception of `shortened_dept` and `department_alias`. shortened dept is addressed when running migration 10, but the setting of `department_alias` needs further investiagion

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/apps/data-pipeline/course-scraper/src/index.ts
+++ b/apps/data-pipeline/course-scraper/src/index.ts
@@ -58,6 +58,15 @@ const unitFormatter = new Intl.NumberFormat("en-US", {
   minimumFractionDigits: 2,
 });
 
+const DEPT_TO_ALIAS: Record<string, string> = {
+  COMPSCI: "CS",
+  EARTHSS: "ESS",
+  "I&C SCI": "ICS",
+  IN4MATX: "INF",
+  ENGRMAE: "MAE",
+  WRITING: "WR",
+};
+
 async function fetchWithDelay(url: string, delayMs = 1000) {
   try {
     logger.info(`Making request to ${url}`);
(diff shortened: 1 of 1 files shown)


apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -57,6 +58,15 @@ const unitFormatter = new Intl.NumberFormat("en-US", {
   minimumFractionDigits: 2,
 });
 
+const DEPT_TO_ALIAS: Record<string, string> = {
```
laggycomputer, line 61: is there a reason this is not `as const`? e.g. type inference becomes wrong or impossible

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -355,60 +329,113 @@ async function scrapeCoursesInDepartment(meta: {
...
-  for (const line of courseText) {
-    if (!line.length) {
-      rawCourses.push([...accumulator]);
-      accumulator = [];
-      continue;
-    }
-    accumulator.push(line);
-  }
+
+  const toCourse = (el: DomElement): typeof course.$inferInsert | null => {
```
laggycomputer, line 358: would prefer this remain a named function like `parseRawCourse` in the before blob, because i don't want to figure out which bindings are available from surrounded scope vs being declared

laggycomputer (changes requested):

apps/data-pipeline/course-scraper/src/index.ts
sanskarm7: made`as const` and as additional layer moved some logic to helper

apps/data-pipeline/course-scraper/src/index.ts
sanskarm7: reorganized logic to use similar `parseCourseBlock`

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -57,6 +58,104 @@ const unitFormatter = new Intl.NumberFormat("en-US", {
...
+  "I&C SCI": "ICS",
+  IN4MATX: "INF",
+  ENGRMAE: "MAE",
+  WRITING: "WR",
+} as const;
+
+type DeptAliasMap = typeof DEPT_TO_ALIAS;
+type DeptCode = keyof DeptAliasMap;
+const getDepartmentAlias = (dept: string): DeptAliasMap[DeptCode] | null =>
+  (DEPT_TO_ALIAS as Record<string, DeptAliasMap[DeptCode]>)[dept] ?? null;
```
laggycomputer, line 73: use stdlib orNull here

laggycomputer (changes requested):
typescript if it was good

laggycomputer (approved):

Issue #200, opened 2025-08-27
reimplement library traffic


sanskarm7:
previous version of library traffic endpoint and scraper was causing `proxy request failed, cannot connect to specified resource` on cloudflare - needs investigation

PR #169, opened 2025-05-21
feat: edited scraper logic to crawl catalogue data(unfinished)

waterkimchi:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code involves a change to the database schema.
- [ ] My code requires a change to the documentation.

PR #204, opened 2025-09-08
feat: library traffic

sanskarm7:
re-implemented version of library-traffic. there is a need to review scraper to validate successful use. the issue with cloudflare spikes must also be investigated.

## Related Issue

#200 

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/graphql/resolvers/library-traffic.ts
+++ b/apps/api/src/graphql/resolvers/library-traffic.ts
@@ -7,7 +7,7 @@ export const libraryTrafficResolvers = {
   Query: {
     libraryTraffic: async (_: unknown, args: { query: unknown }, { db }: GraphQLContext) => {
       const service = new LibraryTrafficService(db);
-      const input = libraryTrafficQuerySchema.parse(args.query);
+      const input = libraryTrafficQuerySchema.parse(args.query ?? {});
       const res = await service.getLibraryTraffic(input);
 
       if (res.length === 0) {


apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,232 @@
...
+
+export async function doScrape(db: ReturnType<typeof database>) {
+  console.log("Starting library traffic scrape.");
+  const locationMeta = await collectLocationMeta();
+
+  const currentTime = new Date();
+  const currentMinute = currentTime.getMinutes();
+
+  for (const [id, meta] of Object.entries(locationMeta)) {
+    const libraryCodeMap = {
```
laggycomputer, line 168: this does not need to be instantiated every time we enter this loop and should be declared once at file scope

laggycomputer (changes requested):
we need to figure out whether emdash or hyphen should be used for comments in this code

LLM moment

apps/api/src/rest/routes/library-traffic.ts
```diff
@@ -0,0 +1,59 @@
...
+const libraryTrafficRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
+
+const libraryTrafficRoute = createRoute({
+  summary: "Retrieve latest library traffic data",
+  operationId: "libraryTraffic",
+  tags: ["Library Traffic"],
+  method: "get",
+  path: "/",
+  request: { query: libraryTrafficQuerySchema },
+  description: "Retrieves the latest traffic data for all floors or a specific floor.",
```
laggycomputer: I think the part about floor(s) is a little confusing for a new reader. Perhaps this description should briefly explain what library traffic is and leave it at that.

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,233 @@
...
+    await db
+      .insert(libraryTraffic)
+      .values([
+        {
+          id: locationData.id,
+          libraryName: meta.libraryName,
+          locationName: meta.locationLabel,
+          trafficCount: locationData.count,
+          trafficPercentage: locationData.percentage,
+          timestamp: new Date(locationData.timestamp),
```
laggycomputer: You should store this `Date` for reuse.

packages/db/src/schema.ts
```diff
@@ -757,6 +758,33 @@ export const apExamReward = pgTable("ap_exam_reward", {
...
 
+export const libraryTraffic = pgTable(
+  "library_traffic",
+  {
+    id: integer("id").primaryKey(),
+    libraryName: varchar("library_name").notNull(),
+    locationName: varchar("location_name").notNull(),
+    trafficCount: integer("traffic_count").notNull(),
+    trafficPercentage: real("traffic_percentage").notNull(),
+    timestamp: timestamp("timestamp").notNull(),
```
laggycomputer: Do you have thoughts on whether this should be named `lastUpdated`/`last_updated` like Websoc instead? I think it's sufficiently clear what a timestamp might mean in this context, but let me know what you think.

laggycomputer (changes requested):
This PR is in a good place and we're close to done.

It really does seem like we spooked the errors into disappearing; the scraper has run perfectly for the last 7 days. I'll shut it down now in anticipation that it will be superseded by this PR being merged.

packages/db/src/schema.ts
estao1: I think timestamp would make more sense; last updated can imply that the library history rows are being updated, which can cause confusion when we implement the historical averaging.

apps/api/src/graphql/schema/library-traffic.ts
```diff
@@ -0,0 +1,19 @@
+export const libraryTrafficSchema = `#graphql
+type LibraryTraffic {
```
laggycomputer: Do you want a cache control directive here?

apps/api/src/schema/library-traffic.ts
```diff
@@ -0,0 +1,47 @@
...
+      description: "Filter results by exact location name",
+    })
+    .optional(),
+});
+
+export const libraryTrafficEntrySchema = z.object({
+  id: z.number().int().nonnegative().openapi({ example: 245 }),
+  libraryName: z.string().openapi({
+    example: "Langson Library",
+    description: "Name of the library that owns this location",
```
laggycomputer: "Owns" is unusual language that may raise some mental questions - I think you should reword using something like "the library which contains"

apps/api/src/schema/library-traffic.ts
```diff
@@ -0,0 +1,47 @@
...
+  trafficPercentage: z.coerce.number().openapi({
+    example: 0.33,
+    description: "Occupancy as a decimal percentage of total capacity (0 to 1)",
+  }),
+  timestamp: z.coerce
+    .date()
+    .transform((d) => d.toISOString())
+    .openapi({
+      example: "2025-01-01T12:34:56.789Z",
+      description: "When the library traffic was recorded (ISO timestamp)",
```
laggycomputer: The correct term for this format is ISO 8601.

apps/api/src/services/library-traffic.ts
```diff
@@ -0,0 +1,35 @@
...
+import { libraryTraffic } from "@packages/db/schema";
+import type { z } from "zod";
+
+type LibraryTrafficServiceInput = z.infer<typeof libraryTrafficQuerySchema>;
+
+export class LibraryTrafficService {
+  constructor(private readonly db: ReturnType<typeof database>) {}
+
+  getLibraryTraffic(input: LibraryTrafficServiceInput) {
+    const queryParams = [];
```
laggycomputer: We typically call an array used for this kind of pattern `conds`. Let me know if there's a compelling reason for this alternative.

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,255 @@
...
+// Helper to sanitize and split raw JS array text content from HTML
+function parseArray(src: string): string[] {
+  return src
+    .split(/[,\n]/)
+    .map((s) => s.trim().replace(/^['"]|['"]$/g, ""))
+    .filter(Boolean);
+}
+
+// Build map of location metadata by ID from the UCI Libraries website
+async function collectLocationMeta(): Promise<Record<string, LocationMeta>> {
```
laggycomputer: Somewhat similarly to what has been added to #169, you should add code comments so that this code can be comfortably read without working through the exact behavior of each line of code, which for scraper code is especially tiring.

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,255 @@
...
+  // Extract parallel arrays:
+  // - locationIds: unique Occuspace IDs for each location (i.e. ['245', '677', '205'])
+  // - floors: corresponding internal DOM codes for each location's container (i.e. ['-GSC-2', '-SL-3', '1'])
+  const locationIds = parseArray(locMatch[1]);
+  const floorCodes = parseArray(floorMatch[1]);
+
+  if (locationIds.length !== floorCodes.length)
+    throw new Error("locationIds and floors arrays are different lengths");
+
+  const codeToLibrary = (code: string): string => {
```
laggycomputer: This isn't a function of anything in this body, so it should be a non-anonymous function at file scope.

packages/db/src/schema.ts
```diff
@@ -757,6 +758,33 @@ export const apExamReward = pgTable("ap_exam_reward", {
...
+  },
+  (table) => [index().on(table.locationName)],
```
laggycomputer: You may also want an index on `libraryName` since queries can be done using that as a filter. Otherwise, such queries will resort to a full table scan (!!!)

laggycomputer (changes requested):

laggycomputer (changes requested):
Do one more check that all the libraries' hours are still the same (I only knew Gateway changed by personal experience), and we should be ready or close to ready

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,267 @@
...
+  return locationMeta;
+}
+
+async function fetchLocation(id: string): Promise<RawRespOK["data"] | null> {
+  const url = `https://www.lib.uci.edu/sites/all/scripts/occuspace.php?id=${id}`;
+  try {
+    // Double-encoded JSON: double parse required
+    const parsedResponse = await fetch(url)
+      .then((r) => r.text())
+      .then(JSON.parse)
```
laggycomputer: Why collect the entire string just to `JSON.parse` it? `Response.json` exists so that it can work on the TCP stream instead of temporarily allocating the entire response body.
```suggestion
      .then((r) => r.json())
```

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (day === 5 && hour >= 8 && hour < 21) return "active";
+    // Sat: Open from 5pm to 9pm
+    if (day === 6 && hour >= 17 && hour < 21) return "active";
+    // Sun: Open from 5pm to 3am the next day
+    if (day === 0 && (hour >= 17 || hour < 3)) return "active";
+
+    // Outside of above hours, the library is closed
+    return "idle";
+  }
+  return "skip";
```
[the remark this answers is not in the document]
laggycomputer: Pretty sure that counts as an error log event for Cloudflare's purposes, so good enough. Hard exception failure is not an option because we need to get whatever data we can without being stuck in a failure loop.

packages/db/src/schema.ts
```diff
@@ -799,7 +799,7 @@ export const libraryTraffic = pgTable(
     trafficPercentage: real("traffic_percentage").notNull(),
     timestamp: timestamp("timestamp").notNull(),
   },
-  (table) => [index().on(table.libraryName), uniqueIndex().on(table.locationName)],
+  (table) => [index().on(table.libraryName, table.locationName)],
```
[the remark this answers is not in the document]
laggycomputer: Also, make sure you re-generate migration files after making any schema changes!

PR #217, opened 2025-09-19
chore: Add minimumReleaseAge

### LINKED ISSUE #215: Set pnpm `minimumReleaseAge`
There were not [one](https://www.upwind.io/feed/npm-supply-chain-attack-massive-compromise-of-debug-chalk-and-16-other-packages) but [two](https://thehackernews.com/2025/09/40-npm-packages-compromised-in-supply.html) rounds of supply chain attacks this month on the NPM registry.

We can set [`minimumReleaseAge`](https://pnpm.io/settings#minimumreleaseage) to prevent such packages from being installed.

A sensible option seems to be 1440 minutes = 1 day.
If we really need a brand new release, we can exempt packages individually using [`minimumReleaseAgeExclude`](https://pnpm.io/settings#minimumreleaseageexclude).

See also https://github.com/icssc/peterportal-client/pull/770.

YumengLiu6044:
## Description
Require packages to be at least a day old for pnpm to consider install them. Prevents instabilities described in #215. After merging this, everyone will need to update their [pnpm version](https://pnpm.io/settings#minimumreleaseage) to 10.16.0 or above. 

This needs a documentation update on the contributor docs.

## Changes

- Added a line in the `pnpm-workspace.yaml` to make sure packages have to be at least 1 day old. Followed [this example](https://github.com/icssc/peterportal-client/pull/770)
- Set pnpm version requirement to v10.16.0 or above to make sure the `minimumReleaseAge` setting works.
- Set a `engine-strict=True` rule for make sure pnpm version is correct.

## Issues

Closes #215


package.json
```diff
@@ -27,5 +27,8 @@
     "prettier-plugin-toml": "2.0.1",
     "tsx": "4.19.2",
     "typescript": "5.6.3"
+  },
+  "engines": {
+    "pnpm": ">=10.16.0"
```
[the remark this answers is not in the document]
YumengLiu6044: Yep, I changed it @laggycomputer . Thanks for the review

laggycomputer (changes requested):
With Actions now permitted for this PR, it looks like the build is now failing specifically because of `pnpm`; can you rectify this?

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
laggycomputer, line 1: What was your basis for adding this? I would imagine that we intend `minimumReleaseAge` to prevent two scenarios:

* Compromise during installation of dependencies on a local machine: `pnpm` always enforces `engines` in this case without this field.
* Compromise of a CI deployment: your PR already specifies the version of `pnpm` to use, so `engines-strict` would not act.

The only scenario where that field would act is if the entire project was somehow installed as a dependency, which doesn't happen. Is there a specific reason why this field is needed? I would prefer to err on the side of not forcing configuration if there's no express reason to do so.

.github/actions/setup-node-and-pnpm/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
       uses: pnpm/action-setup@v2
       with:
         run_install: false
-        version: 9
+        version: 10.16.0
```
laggycomputer, line 20: There's been a new `pnpm` release `v10.17.0` which adds slightly more functionality to the `minimumReleaseAgeExclude` setting. On the off chance that we use that in the future, let's add that minor bump with this major bump, since it has no extra migration cost.

This release came out very recently (less than 60 hours ago), so I missed it initially too.

laggycomputer (changes requested):

laggycomputer:
Also, don't worry about Actions now failing - Actions runs on your PR are not provided some environment variables which are necessary to create a deployment, since your PR comes from your fork. As long as `pnpm` is set up and packages are installed successfully, I'm reasonably certain that the deployment would have worked.

.npmrc
YumengLiu6044: I guess it's not necessary. I will remove this

YumengLiu6044:
Alright I made those changes @laggycomputer

laggycomputer (changes requested):
Thanks for your patience. There is another warning being displayed in install logs about approved build scripts. These are safe to enable, and that decision persists in `onlyBuiltDependencies`. Can you whitelist every build script currently being blocked?

sanskarm7:
Hi Yumeng,

Thank you for taking the initiative on this! We’re currently having some internal discussions about upgrading to pnpm 10. Since many other ICSSC projects are still on pnpm 9, moving the API to pnpm 10 right now could cause workflow inconsistencies for developers.

With that in mind, we’d like to hold off on merging this change until timelines across projects are more aligned.

In the meantime, I’d highly encourage you to check out one of the other “good first tasks". Your contributions there would be really valuable.

Thanks again for your effort!

Issue #94, opened 2025-01-24
Year schema check is incorrect


laggycomputer:
The current year schema fails to catch any missing value, because the if check was written incorrectly ([code](https://github.com/icssc/anteater-api/blob/main/apps/api/src/schema/lib/year.ts#L9)). The right hand of the strict inequality falls back to `"undefined"` and only this is checked. The author likely intended all three to be checked.

We should either correct this or remove that refinement step.

andrew-wang0:
Looks like this is also causing a mismatch between the openapi spec.

Below: year appears to be optional, but is actually required
![Image](https://github.com/user-attachments/assets/08106343-116f-41f8-a6d6-d91e23c946ba)

![Image](https://github.com/user-attachments/assets/a60ba8e3-e872-4160-9859-04172e01c3eb)

hpathanjaly:
Testing a fix for this issue right now

PR #226, opened 2025-10-06
fix: explicit check for missing value in year schema

hpathanjaly:
## Description
Solved issue #94 by explicitly comparing val to the empty string, null and undefined

## How Has This Been Tested?
Ran null, undefined, and empty string through localhost. Results in screenshots

## Screenshots
image
image
image

## Types of Changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/lib/year.ts
+++ b/apps/api/src/schema/lib/year.ts
@@ -6,10 +6,10 @@ import { z } from "@hono/zod-openapi";
 
 export const yearSchema = z.coerce
   .string()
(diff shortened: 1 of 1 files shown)


laggycomputer:
> is there a reason for not type annotating in TypeScript in regards to speed or something? Or is it to avoid being too verbose in this case?
The reason is both fragility and verbosity. Speed is not relevant since types are stripped before runtime interpretation.

hpathanjaly:
push made to remove empty string check

laggycomputer:
Thanks. Just the type hints now.

PR #235, opened 2025-10-08
feat: sample program scraper

aadi-shanker:


apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,590 @@
...
+  }[],
+) {
+  if (!scrapedPrograms.length) {
+    logger.info("No sample programs to store.");
+    return;
+  }
+
+  logger.info("Fetching sample programs from database...");
+
+  // Prepare parent table rows (catalog_program)
```
laggycomputer: This is not the name of this table anymore.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,590 @@
...
+              });
+              break;
+            }
+            currentElement = currentElement.next();
+          }
+        }
+      });
+
+      // Look for <dl class="sc_footnotes">
+      if (variationNotes.length === 0) {
```
laggycomputer: Comment here to make it clear that this is a second-priority fallback (it is, right?)

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,590 @@
...
+      }
+
+      // Look for <ol> after table
+      if (variationNotes.length === 0) {
+        sampleProgramContainer.find("ol").each((_i, ol_el) => {
+          const prevElement = $(ol_el).prev();
+          if (
+            prevElement.is("table") ||
+            prevElement.is("h6") ||
+            (prevElement.is("p") && prevElement.text().toLowerCase().includes("notes"))
```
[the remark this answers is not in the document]
aadi-shanker: Yeah this also answers the prev comment because the notes varied so much I added 2 extra edge cases just as a final fallbacks in the case it happens. Just to put it in as defensive code but here is what i documented with my findings.

Strategy 1: "NOTES:" Heading + Paragraphs/Lists
1. Computer Science, B.S.
2. Business Information Management

Strategy 2: <dl class="sc_footnotes">
1. Civil Engineering (multi-variation programs)
2. Psychology variations

Strategy 3: Paragraphs with * or numbers
Added as fallback (no specific example documented)

Strategy 4: `<ol>` after table
Added as final fallback (no specific example documented)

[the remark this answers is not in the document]
aadi-shanker: Ran a test just to see if it was catching anything and starategy 4 didnt so i went ahead and deleted it but strategy 3 had 3 entineering programs notes caught

packages/db/src/schema.ts
```diff
@@ -145,6 +145,37 @@ export type APCoursesGrantedTree =
...
+// CHILD TABLE: sample_program_variation
+// Stores each variation's course data (multiple rows per program)
+export const sampleProgramVariation = pgTable("sample_program_variation", {
+  id: uuid("id").primaryKey().defaultRandom(),
+  programId: varchar("program_id")
+    .notNull()
+    .references(() => catalogProgram.id, { onDelete: "cascade" }),
+  label: varchar("label"),
+  sampleProgram: json("sample_program").$type<SampleProgramEntry[]>().notNull(),
+  variationNotes: varchar("variation_notes").array().default(sql`ARRAY[]::VARCHAR[]`),
```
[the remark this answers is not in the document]
aadi-shanker: In regards to label single tables dont have or show a label so its null. and for variation notes if there is no notes i just keep it blank which is the same with single tables.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,590 @@
...
+
+const MAX_DELAY_MS = 8_000;
+
+const HEADERS_INIT = {
+  Connection: "keep-alive",
+};
+
+type SampleYear = {
+  year: AcademicYearType;
+  curriculum: string[][];
```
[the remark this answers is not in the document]
aadi-shanker: Oh yeah so it is a intermediate/raw format from scraping, before transformation. For example

// Step 1: Raw scraped data
SampleYear = {
  year: "Freshman",
  curriculum: [
    ["MATH 2A", "MATH 2B", "MATH 2D"],  // Row 1
    ["WRITING 39A", "WRITING 39B", "WRITING 39C"]  // Row 2
  ]
}

// Step 2: Transformed
SampleProgramEntry = {
  year: "Freshman",
  fall: ["MATH 2A", "WRITING 39A"],
  winter: ["MATH 2B", "WRITING 39B"],
  spring: ["MATH 2D", "WRITING 39C"]
}

what i will do is just rename it for more specificity then add a comment explaining the purpose

apps/api/src/schema/programs.ts
```diff
@@ -282,3 +289,67 @@ export const ugradRequirementsResponseSchema = z.object({
...
+  spring: z.array(z.string()).openapi({
+    description: "Courses recommended for Spring term",
+    example: ["I&C SCI 33", "IN4MATX 43", "WRITING 60"],
+  }),
+});
+
+export const sampleProgramVariationSchema = z.object({
+  label: z.string().optional().openapi({
+    description:
+      "Label describing this variation (empty string for programs with single variation)",
```
laggycomputer: I don't think we should explain why labels may be missing because it doesn't give much benefit and we shouldn't make promises that we can't keep. Users can draw an inference about when labels tend to be missing on their own time.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,570 @@
...
+  const variationRows = scrapedPrograms.flatMap((program) =>
+    program.variations.map((variation) => ({
+      programId: program.majorId,
+      label: orNull(variation.label),
+      sampleProgram: variation.sampleProgram,
+      variationNotes: variation.variationNotes,
+    })),
+  );
+
+  // Extract program IDs for reuse
```
laggycomputer: Same here. This is pretty obvious.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,570 @@
...
+
+  if (!readlineSync.keyInYNStrict("Is this ok")) {
+    logger.error("Cancelling sample program update.");
+    return;
+  }
+
+  await db.transaction(async (tx) => {
+    await tx.delete(catalogProgram).where(inArray(catalogProgram.id, programIds));
+
+    // Insert parent rows
```
laggycomputer: Same with this pair of comments.

laggycomputer (changes requested):
The comments in the scraper make the code much more readable, since we no longer have to figure out what each block of code means and can just see what it does at a moderately low level. Good job.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,570 @@
...
+        programPaths.push(href);
+      }
+    });
+  }
+  return Array.from(new Set(programPaths));
+}
+
+/*
+ * Extracts a unique ID from the catalogue URL
+ * Example: https://catalogue.uci.edu/.../computerscience_bs/ → "computerscience_bs"
```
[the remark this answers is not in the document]
aadi-shanker: Yeah LLM wrote that just asked it for any places it recommends comments gave me that will fix my fault

packages/db/src/schema.ts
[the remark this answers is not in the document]
aadi-shanker: Done added not null to the variation notes that way it is completley consistent with the rest of the columns and it would be an empty array

apps/api/src/services/programs.ts
```diff
@@ -139,4 +142,50 @@ export class ProgramsService {
...
+          variations: sql`
+            COALESCE(
+              json_agg(
+                json_build_object(
+                  'label', ${sampleProgramVariation.label},
+                  'sampleProgram', ${sampleProgramVariation.sampleProgram},
+                  'variationNotes', ${sampleProgramVariation.variationNotes}
+                )
+                ORDER BY ${sampleProgramVariation.id}
+              ) FILTER (WHERE ${sampleProgramVariation.id} IS NOT NULL),
```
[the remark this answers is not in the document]
aadi-shanker: Yeah my thought process is we can just exclude the nulls before aggregation then we can just aggregate only valid rows which I think is more effecient just from a speaking standpoint versus manually checking and aggregating everything

apps/api/src/rest/routes/catalogue.ts
```diff
@@ -0,0 +1,67 @@
...
+  "*",
+  productionCache({ cacheName: "anteater-api", cacheControl: "max-age=86400" }),
+);
+
+catalogueRouter.openapi(sampleProgramsRoute, async (c) => {
+  const query = c.req.valid("query");
+  const service = new ProgramsService(database(c.env.DB.connectionString));
+  const res = await service.getSamplePrograms(query);
+
+  if (query?.id && !res.length) {
```
laggycomputer: The fact that you check for an ID here and not in GraphQL suggests that you intended them to have parity. You should decide on the behavior.

apps/api/src/schema/catalogue.ts
```diff
@@ -0,0 +1,105 @@
+import { z } from "@hono/zod-openapi";
+
+export const sampleProgramsQuerySchema = z.object({
+  id: z.string().optional().openapi({
+    description: "Filter sample programs by program ID and returns all programs if omitted.",
+    example: "computerscience_bs",
+  }),
+});
+
+export const standingyearEnum = z.enum(["Freshman", "Sophomore", "Junior", "Senior"]);
```
laggycomputer: It could be better to outline that array for posterity.

apps/api/src/schema/catalogue.ts
```diff
@@ -0,0 +1,105 @@
...
+        { type: "courseId", value: "I&CSCI6B" },
+        { type: "courseId", value: "WRITING60" },
+      ],
+    },
+  });
+
+export const sampleProgramVariationSchema = z.object({
+  label: z.string().nullable().openapi({
+    description:
+      "Label describing this variation of the multiple sample programs available for this program",
```
laggycomputer: I'm not sure we have to tie the existence of a label to the fact that there are multiple variations. I know that's actually the case, but each of these variations should be valid in its own right without concern for what's going on around it. For that reason, you should just describe this field without reference to other variations (including whether they exist or not).

apps/api/src/schema/catalogue.ts
```diff
@@ -0,0 +1,105 @@
...
+  id: z.string().openapi({
+    description: "Stable ID for this sample program record",
+    example: "computerscience_bs",
+  }),
+  programName: z.string().openapi({
+    description: "Program name of this sample program",
+    example: "Computer Science, B.S.",
+  }),
+  variations: z.array(sampleProgramVariationSchema).openapi({
+    description: "Array of program variations. Programs with single variation have empty label.",
```
laggycomputer: As I wrote in the previous review, and for the same reasons above, this second sentence should not be present.

apps/api/src/schema/catalogue.ts
```diff
@@ -0,0 +1,91 @@
...
+    type: z.literal("courseId"),
+    value: z.string().openapi({
+      description: "A string known to be a valid course ID",
+      example: "I&CSCI31",
+    }),
+  }),
+  z.object({
+    type: z.literal("unknown"),
+    value: z.string().openapi({
+      description: "Text that could not be validated (e.g., general education requirements)",
```
laggycomputer: This documentation might give users the wrong idea if we attempt to parse GE requirement indicators later. In fact, those seem like somewhat low-hanging fruit, so that might be the next thing we add! I don't think the "e.g." part here is necessary, because the example probably explains it well enough.

apps/api/src/schema/catalogue.ts
```diff
@@ -0,0 +1,91 @@
...
+export const sampleProgramsQuerySchema = z.object({
+  id: z.string().optional().openapi({
+    description: "Filter sample programs by program ID and returns all programs if omitted.",
+    example: "computerscience_bs",
+  }),
+});
+
+const STANDING_YEARS = ["Freshman", "Sophomore", "Junior", "Senior"] as const;
+
+export const standingyearEnum = z.enum(STANDING_YEARS);
```
laggycomputer: This isn't correct `camelCase` if you meant "standing" and "year" to be different years.

I also think the word "enum" isn't quite normative in this codebase; maybe "schema" is more appropriate.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+  }
+}
+
+/*
+  Transforms course code strings to course IDs by looking them up in the database.
+  Falls back to original string if course not found.
+ */
+async function transformCourseCodesToIds(
+  db: ReturnType<typeof database>,
+  scrapedPrograms: {
```
laggycomputer: Whoa! These inline types are a little unwieldy. I think they would look better outlined, either just above this function or with the other types at the top of the file.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+
+  const transformed: {
+    majorId: string;
+    programName: string;
+    variations: { label?: string; sampleProgram: SampleProgramEntry[]; variationNotes: string[] }[];
+  }[] = [];
+
+  for (const program of scrapedPrograms) {
+    const transformedVariations: {
+      label?: string;
```
laggycomputer: And you're forced to state this one twice too.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+      const transformedSampleProgram: SampleProgramEntry[] = [];
+
+      for (const year of variation.sampleProgram) {
+        // Transform each term's courses in parallel
+        const [fall, winter, spring] = await Promise.all([
+          Promise.all(
+            year.fall.map(async (code) => {
+              const id = await lookupCourseId(db, code);
+              return id
+                ? { type: "courseId" as const, value: id }
```
laggycomputer: This code being repeated three times is cumbersome and error-prone. I think a separate helper is warranted which operates on a single code or array of codes, which can be called three times.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+    });
+  }
+
+  logger.info("Course code transformation complete");
+  return transformed;
+}
+
+async function storeSampleProgramsInDB(
+  db: ReturnType<typeof database>,
+  scrapedPrograms: {
```
laggycomputer: This type looks similar to the above. Hopefully it can be outlined.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+    .select()
+    .from(catalogProgram)
+    .where(inArray(catalogProgram.id, programIds));
+
+  const existingVariations = await db
+    .select()
+    .from(sampleProgramVariation)
+    .where(inArray(sampleProgramVariation.programId, programIds));
+
+  // Combine into unified structure for comparison
```
laggycomputer: I think this comment is unnecessary.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+  const dbData = { catalogPrograms: existingCatalogPrograms, variations: existingVariations };
+  const scrapedData = { catalogPrograms: catalogueRows, variations: variationRows };
+
+  // Sort both structures
```
laggycomputer: I think the variable name and actual code make it clear that this is a sort, so this comment is not needed.

laggycomputer (changes requested):

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -0,0 +1,715 @@
...
+
+    await tx.insert(catalogProgram).values(catalogueRows);
+
+    await tx.insert(sampleProgramVariation).values(variationRows);
+  });
+
+  logger.info(`Successfully stored ${catalogueRows.length} sample programs`);
+}
+
+// HELPER: Parse a single table and return its data
```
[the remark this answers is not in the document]
aadi-shanker: OH I do label comments sometimes just for an organizing purpose seperating helpers from actual functions in the case the helper function is as detailed but yes I shoulda have more in depth comment forgot to update my bad.

apps/data-pipeline/sample-program-scraper/src/index.ts
```diff
@@ -199,73 +229,42 @@ async function lookupCourseId(
...
 }
 
+/*
+  Validates a single course code and returns a discriminated union.
+  Returns courseId type if found in database, unknown type otherwise.
+*/
+async function validateCourseCode(
+  db: ReturnType<typeof database>,
+  code: string,
+): Promise<{ type: "courseId"; value: string } | { type: "unknown"; value: string }> {
```
laggycomputer: I think this should be outlined too. It's one of the most likely points to be changed in the future.

laggycomputer (changes requested):
Definitely cleaner now.

laggycomputer (approved):
@sanskarm7 go test since i didn't

sanskarm7 (approved):
testing looks good!

great work aadi!! 🎉

aadi-shanker:
> testing looks good!
> 
> great work aadi!! 🎉

Bang!!!! thanks for all your help @laggycomputer

Issue #164, opened 2025-05-18
AP exam importer should clear `ap_exam` table


laggycomputer:
...or do an `ON CONFLICT` `SET`. That way clearing the table will not be necessary to run the importer.

The other two tables do not need to be cleared because they use random UUID primary keys.

PR #237, opened 2025-10-13
Fix: update SQL to upsert data in APExamRewards instead of clearing and filling table.

Bennables:
Current APexamimporter requires admins to delete the whole table of APExamRewards table before filling it again. This pull request updates the code to upsert it instead, saving computational time

## Description

Fix: Apexamimporter upserts data instead of trying to replace and causing errors.

I used onConflictDoUpdate, which updates the data when trying to insert repeated entries for AP exams.
I removed rows that weren't referenced by apexamtoreward in apexamreward.

## Related Issue

#164 

## Motivation and Context

This change is required because it allows you to run APExamImporter's index without clearing the Apexam in the database.

## How Has This Been Tested?

This remains true in the most recent commit. 
1. prior to this change, there was an error because there was a conflict(this is why we wiped the table). Now, if you run the script again, there is no error. Further testing reveals that the id of the apexamToReward table will stay the same indicating the update worked. 
I deleted some rows and imported again and the ids changed, so that means the rows are really getting updated.
2. this removes rows from apExamReward not mentioned in apExamToReward. The table shrank by ~350 entries and all rewards can be found on the table now.

## Types of changes

## Checklist:

### DIFF
--- a/apps/key-manager/docker-compose.yml
+++ b/apps/key-manager/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       context: src/db
       dockerfile: Dockerfile
     ports:
-      - "5432:5432"
+      - "5433:5432"
     environment:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres


apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,11 +50,27 @@ async function main() {
           .then((rows) => rows[0]);
 
         for (const score of reward.acceptableScores) {
-          await tx.insert(apExamToReward).values({ examId: fullName, score, reward: rewardId });
+          await tx
+            .insert(apExamToReward)
+            .values({ examId: fullName, score, reward: rewardId })
+            // assuming that reward is where the conflict is going to be
```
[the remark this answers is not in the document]
Bennables: yeah, this was for when I was initially reading the code
it has been removed

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,11 +50,27 @@ async function main() {
...
+            .onConflictDoUpdate({
+              target: [apExamToReward.examId, apExamToReward.score],
+              set: conflictUpdateSetAllCols(apExamToReward),
+            });
+          // .onConflictDoNothing({target:[apExamToReward.examId, apExamToReward.score]})
```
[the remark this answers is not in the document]
Bennables: this has been removed because it is not used

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,12 +55,29 @@ async function main() {
...
+            .onConflictDoUpdate({
+              target: [apExamToReward.examId, apExamToReward.score],
+              set: conflictUpdateSetAllCols(apExamToReward),
+            });
         }
       }
     }
   });
-
+  //remove apExamRewards that don't exist in the junction table(ap_exam_to_reward)
```
laggycomputer: This comment is unnecessary, in my opinion. If retained, use `// ` (note the extra space).

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,12 +55,29 @@ async function main() {
...
+  //done using correlated subquery
+  await db
+    .delete(apExamReward)
+    .where(
+      notExists(
+        db
+          .select({ one: sql`1` })
```
laggycomputer: Analysis in query planners reveals that it does not matter what you place in this `select` because it does not affect the query. In this case, it only adds some degree of uncertainty for code readers because raw SQL is being used. The default `.select()` with no arguments is sufficient.

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,12 +55,29 @@ async function main() {
...
       }
     }
   });
-
+  //remove apExamRewards that don't exist in the junction table(ap_exam_to_reward)
+  //done using correlated subquery
+  await db
+    .delete(apExamReward)
+    .where(
+      notExists(
```
laggycomputer: You should investigate whether something of the form `DELETE FROM a WHERE a.pkey IN (SELECT a.pkey FROM a LEFT JOIN b ON ... WHERE b.pkey IS NULL)` is faster, for table `a` having a one-to-many relationship with `b`. I'm not sure myself, but I suspect the join-based proposal is faster. My intuition is that many `EXISTS`/`NOT EXISTS` on one row at a time is not as likely to optimize.

packages/db/src/index.ts
```diff
@@ -1,7 +1,9 @@
...
+import * as drizzleHelpers from "drizzle-orm";
 import { drizzle } from "drizzle-orm/postgres-js";
 
 export const database = <TSchema extends Record<string, unknown>>(
   url: string,
-  params?: DrizzleConfig<TSchema>,
+  params?: drizzleHelpers.DrizzleConfig<TSchema>,
 ) => drizzle({ connection: { url, max: 1 }, ...params });
+
+export const { eq, notExists, sql } = drizzleHelpers;
```
laggycomputer: This is a fragile and somewhat arbitrary way to gain access to these primitives. A wildcard import might be best.

Better yet, read the `packages/db/src/drizzle.ts` file. There may be a better way involving that file to gain access to Drizzle functions.

apps/data-pipeline/ap-exams-importer/src/index.ts
Bennables: resolved

apps/data-pipeline/ap-exams-importer/src/index.ts
Bennables: I have reworked this whole sectoin of code

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,12 +56,24 @@ async function main() {
...
+            .onConflictDoUpdate({
+              target: [apExamToReward.examId, apExamToReward.score],
+              set: conflictUpdateSetAllCols(apExamToReward),
+            });
         }
       }
     }
   });
 
+  const arr = (await db.select({ reward: apExamToReward.reward }).from(apExamToReward)).map(
```
laggycomputer: This query does not need to be collected into an array in JavaScript, which incurs an additional round of network and memory cost. This should only be done if some processing step is actually more performant despite that downside or excessively complex in pure SQL. Our case here is neither and a subquery should be used.

apps/key-manager/.env.example
```diff
@@ -1,6 +1,6 @@
 AUTH_SECRET=
 
-USERS_DB_URL=postgres://postgres:postgres@localhost:5433/anteater_api_users
+USERS_DB_URL=postgres://postgres:postgres@localhost:5432/anteater_api_users
```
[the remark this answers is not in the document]
Bennables: misclicked my ctrl-f

apps/data-pipeline/ap-exams-importer/src/index.ts
```diff
@@ -50,12 +56,24 @@ async function main() {
...
+    (row) => row.reward,
+  );
+
+  db.delete(apExamReward).where(notInArray(apExamReward.id, arr));
```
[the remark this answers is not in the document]
Bennables: this has been awaited

apps/data-pipeline/ap-exams-importer/src/index.ts
Bennables: removed and put into a subquery

pnpm-lock.yaml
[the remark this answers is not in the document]
Bennables: yeah, I changed it back

PR #252, opened 2025-11-03
feat: /websoc/departments

### LINKED ISSUE #236: Endpoint to list all departments
There is currently no way to get all the (recent or relevant) departments, leading to non-ideal patterns like AntAlmanac's [hardcoded departments list](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/constants.ts).

It would be helpful to have an endpoint that, given a start term, retrieves all departments since that term.

## Expected Behavior

The endpoint can be called with a start term and responds with all departments that have had courses offered since that term.
- The start term parameter is important because some departments have been unused for decades and are entirely irrelevant to the vast majority of users
- The response should include their human-readable name (e.g. `BIO SCI` is named `Biological Sciences`) as that will very likely be desired by users (see context below)
- It would also be helpful to have the option to include a list of all course IDs for each of the departments (see context below)

## Context

As mentioned above, projects that display a list of departments are currently forced to hardcode their departments. [AntAlmanac's hardcoded list](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/constants.ts) for their search menu is a prevalent ICSSC example that would benefit from this, and [PeterPortal's upcoming search filters](https://github.com/icssc/peterportal-client/pull/800) is in the same situation.

PeterPortal is also considering a "catalogue view" in which users can select a subset of departments and view every course ID of each selected department. Fetching that data would be made much more straightforward if it was possible to get all course IDs arranged by department, as described above, to prevent complex solutions like paginating over the courses endpoint.

laggycomputer:
## Description

A new endpoint and GraphQL resource are introduced to get a list of all valid department codes and names.
Per the linked issue, optional parameters also allow for restriction through a "since" point in time.
Departments which have not been seen in WebSoc since that year or year and term will not be returned.

## Related Issue

Closes #236.

## Motivation and Context

the years of hardcoding depts are over

## How Has This Been Tested?

curl

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/websoc.ts
+++ b/apps/api/src/schema/websoc.ts
@@ -311,7 +311,7 @@ export const websocDepartmentsQuerySchema = z
 
 export const websocDepartmentsResponseSchema = z
   .object({
-    deptCode: z.string(),
-    deptName: z.string(),
+    deptCode: z.string().openapi({ example: "BIO SCI" }),
+    deptName: z.string().openapi({ example: "Biological Sciences" }),
   })
   .array();


apps/api/src/schema/websoc.ts
```diff
@@ -292,3 +292,25 @@ export const websocTermResponseSchema = z.object({
...
 });
+
+export const websocDepartmentsQuerySchema = z
+  .object({
+    sinceYear: yearSchema.optional().openapi({
+      description: "If provided, departments must have appeared at least once not before this year",
+    }),
+    sinceQuarter: z.enum(terms).optional().openapi({
+      description:
+        "If provided with `sinceYear`, departments must appear at least once not before this quarter in that year",
```
sanskarm7: similarly here

sanskarm7 (changes requested):
this also provides data for 2026, specifically for Winter, which is expected.

although not directly a logic problem in the scope of the endpoint, i wanted to get your thoughts on changing how we frame the documentation to show that the information for new quarters is available - even if it's in the "future". not sure if this is self explanatory for those who are using this endpoint but some level of observability may help.

one thing i propose is `since` fields in the return which would show changes as they occur quarter by quarter. if you feel this is unnecessary though, i don't feel too strongly on this.. just looking for ways to increase observability as mentioned earlier.

apps/api/src/schema/websoc.ts
laggycomputer: missed a word too lol

sanskarm7 (approved):
lgtm

Issue #249, opened 2025-11-02
[websoc] SQL junction table column name isn't right


laggycomputer:
https://github.com/icssc/anteater-api/blob/3e63d1eb0fbde7830959d3170a18ccb581239787/packages/db/src/schema.ts#L413-L415

Here, we define a column, give it the name `meetingId` in Drizzle, and say it refers to the ID of a meeting of a Websoc section. However, we name the column `section_id` in SQL. This doesn't make sense, since this column doesn't refer to a section.

PR #260, opened 2025-11-16
fix: section_id column to meeting_id for clarity

natran17:
The column references websocSectionMeeting.id (a meeting ID), but was incorrectly named section_id.

## Related Issue

https://github.com/icssc/anteater-api/issues/249

## Motivation and Context

The previous naming was confusing and misleading for readers. Now it is more intuitive.

## How Has This Been Tested?

Only column name was changed. It had no effect on the code, so the code runs fine just like before.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -15,7 +15,7 @@
     "@graphql-tools/merge": "9.0.8",
     "@graphql-yoga/plugin-response-cache": "3.10.0",
     "@hono-rate-limiter/cloudflare": "0.2.1",
-    "@hono/zod-openapi": "0.16.4",
+    "@hono/zod-openapi": "1.1.5",
     "@packages/db": "workspace:*",
     "@packages/stdlib": "workspace:*",
     "@scalar/hono-api-reference": "0.5.158",
@@ -23,7 +23,7 @@
     "graphql-yoga": "5.8.0",
     "hono": "4.6.8",
     "hono-rate-limiter": "0.4.0",
-    "zod": "3.23.8"
+    "zod": "4.1.13"
   },
   "devDependencies": {
     "@packages/key-types": "workspace:*",
--- a/apps/api/src/schema/ap-exams.ts
+++ b/apps/api/src/schema/ap-exams.ts
@@ -55,7 +55,7 @@ export const apExamsRewardSchema = z.object({
     description:
       "The number of units granted as generic elective credit (but not from any course) for this reward",
   }),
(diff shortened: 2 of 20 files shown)


laggycomputer (changes requested):
Thanks for the PR.
- This does not require a documentation change, since it changes an internal implementation detail.
- Currently, this PR does nothing because there is no migration. More information [here](https://orm.drizzle.team/docs/migrations).

packages/db/src/migrate.ts
```diff
@@ -1,3 +1,4 @@
+import "dotenv/config";
```
laggycomputer: We shouldn't need this. The scripts in `package.json` files wrap all code execution in `dotenv`, so the code should not need to do anything to set up environment variables.

laggycomputer (changes requested):
Thanks for the patch. The idea looks good, but you'll have to regenerate the migration file because a migration sharing ID 19 has been merged into the codebase since this PR was opened. Migrations are sequential so this ID overlap must be resolved.

Fixing this is a little troublesome:
* Rectify your branch so that the `db` folder exactly matches what's currently on our `main` branch, with no added or deleted files.
* Make the change to the column name again and generate a new migration. Make sure you name the migration per my comments.
* Commit the result.

This is a somewhat annoying artefact of the way Drizzle works. I'm happy to guide you through this (contact me/us on Discord) if you need help, because I see this as an undue burden on first-time contributors.

natran17:
Sorry for not notifying your about the updated PR. I'm glad you were able to find it and thank you for all the helpful comments. I created a new migrate ID. Let me know if everything looks good. Thank you!

package-lock.json
laggycomputer: This file looks like you attempted to use `npm` as package manager. Per our [contributor docs](https://docs.icssc.club/docs/contributor/anteaterapi/setup) and codebase, we use `pnpm`. You should use that package manager and revert changes to package-manager related files. Those should not be necessary for this PR.

laggycomputer (changes requested):
The migration stuff looks fine now. Looks like some other artefacts got mixed in. Clear those out and we should be good.

natran17:
Thanks for pointing those out. I've corrected them now.

laggycomputer (approved):
Alright, this is good now. Thanks for the patch!

laggycomputer (changes requested):
Seconding the above, also please follow the Conventional Commits spec in the PR title since this will become the commit message.

natran17:
Thanks for all the thorough reviews and the patience. Appreciate your help.

Issue #256, opened 2025-11-13
Add "until" to /websoc/departments


laggycomputer:
Blocked by #252.

We'd like to add an "until" pair of parameters as a complement to the "since" pair, so that collection of department entities for the result can be restricted in the future direction as well.
This enables other applications like comparison of department offerings between different terms (if "until" and "since" are set to the same term) or preventing departments from terms in the future from being included.

PR #261, opened 2025-11-16
feat: Add websoc departments until

limlindsayj:
## Description
Add "untilYear" and "untilQuarter" parameters to websoc departments endpoint.

## Related Issue
Issue #256 

## Motivation and Context
"We'd like to add an "until" pair of parameters as a complement to the "since" pair, so that collection of department entities for the result can be restricted in the future direction as well.
This enables other applications like comparison of department offerings between different terms (if "until" and "since" are set to the same term) or preventing departments from terms in the future from being included."

## How Has This Been Tested?
Ran endpoint.

## Types of changes

## Checklist:


sanskarm7 (approved):
tests well!

fantastic work lindsay :D

Issue #172, opened 2025-05-23
Add some indication of cross-listed courses


Awesome-E:
Originally received via PeterPortal Feedback Form:

> Improved handling of code-share classes. A simple hyperlink that says this class may be offered under a different name is enough. e.g., ENGR30 and ENGRMAE30, is the same class. Same thing with ENGR 150 and ENGRMAE150. This causes major headaches when planning some GE Courses, which are offered under different codes by quarter (History 15C vs. Asian American Studies 50). It may appear as if a class isn't offered when in reality it is. Neither the catalog nor advising makes this clear, so this could be a chance for PeterPortal to provide much-needed clarity on this issue.

PR #269, opened 2025-11-23
feat: course alternate names

ParzivalPerhaps:
## Description

Addressing #172, adds an `aliases` column to the `course_view` materialized view that finds courses with matching descriptions (which seems to be the most effective way to find courses under different names) and collects their course IDs as a string array.

It applies to all course requests, in the case of no aliases it returns an empty array, otherwise, it returns a list of course IDs.

### Disclaimer
There is an existing `same_as` column which does not seem to work consistently or provide a robust solution, so it seemed more straightforward to create a new column. We can leave it as is to avoid potentially breaking changes, but I'd recommend deprecating it.

## Related Issue

#172 

## Motivation and Context

This is useful information for PeterPortal/AntAlmanac to annotate course entries since UCI enjoys renaming and relabeling courses for some reason.

## How Has This Been Tested?

In a locally run environment.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/graphql/schema/courses.ts
+++ b/apps/api/src/graphql/schema/courses.ts
@@ -33,6 +33,7 @@ type Course @cacheControl(maxAge: 86400) {
     geList: [String!]!
     geText: String!
     terms: [String!]!
+    aliases: [String!]!
 }
 
 type CoursesByCursor {


laggycomputer (changes requested):
This heuristic makes basic sense but I'd like more proof that this is necessary. You mention WebSoc `same_as` is insufficient, but it captures not only correspondence between courses but their individual sections as well. It captures the three examples from the feedback ticket along with the correspondence between HISTORY 15C and SOCSCI 78A not originally in the ticket.

* What exactly is the name of this relation? It's been referred to as "same as" (because WebSoc displays this), "cross-listed courses" (a more colloquial term), and you're now calling them "aliases" (alternate casual term?).
* Why do we need a new heuristic? Are there examples of this correspondence which are not captured by the "same as" relation scraped from WebSoc, requiring us to impute this data ourselves?

Issue #266, opened 2025-11-18
WebSoc `is_cancelled` column is not correct


laggycomputer:
This column is governed by a `GENERATED ALWAYS AS` which checks for `websoc_section.section_comment LIKE '*** CANCELLED ***%'`. However, for the Winter 2026 section with code 34021 (COMPSCI 116), we see the following comment:

```

			    <p>***  CANCELLED  *** Course Code 34021</p>
			    
                
```

Which clearly doesn't obey what we expect. When we fix this, I think we might as well also allow the term to be spelled "canceled". The details of how precisely this should be re-done are up to the assignee.

PR #272, opened 2025-12-01
fix: update websoc is_cancelled column to handle websoc data formatting

### LINKED ISSUE #168: [WebSoc] Provide indication of cancelled courses
Currently, WebSoc sections that have been cancelled have no explicit indication thereto in the section response, except through the section comment. We already have a [generated column](https://github.com/icssc/anteater-api/blob/main/packages/db/src/schema.ts#L332-L336) whose value is informed by the section comment for filtering purposes. It would be trivial to add this flag to the section response; the issue is how to convey this in the response.

The naive solution is of course to add an `isCancelled` field to the section response. However, given that sections retain their pre-cancellation status (sample query: https://anteaterapi.com/v2/rest/websoc?year=2025&quarter=Spring&department=COMPSCI&sectionCodes=34671) which is now semantically meaningless, @Awesome-E proposed that cancelled sections should simply have their status set to "Cancelled" or similar.

I agree that replacing the status would make the most sense, but it would also result in loss of data as the original status is no longer conveyed. However, it can be argued that the original status is no longer relevant once a section is cancelled. Would love to hear your opinions on the matter.

cc: @icssc/api, @icssc/antalmanac, @icssc/peterportal

iAmbrosial:
## Description

The pattern matching logic for the `is_cancelled` generated column in the `websoc_section` table was fixed to correctly identify cancelled sections by parsing the section comment. Additionally, the term has been allowed to either be spelled "cancelled" or "canceled". 

The isCancelled boolean field has also been added to the websocSectionSchema Zod Schema with OpenAPI docs, as well as the WebsocSection GraphQL type with documentation. This provides an explicit indicator of cancelled course sections to end users, and takes precedence over the existing `status` field in the API response.

## Related Issues

#266 

Also resolves #168 

## Motivation and Context

The current logic for generating the `is_cancelled` column is incorrect, leading to misidentification of cancelled class sections. The WebSoc API returns section comments with HTML formatting and extra whitespace, and this did not match the expected pattern matching logic controlling the generated column. This needs to be fixed so that accurate data regarding cancelled classes–both past and future–are output when queried. 

We want the end user to be able to access the `isCancelled` field that indicates whether or not the corresponding class section has been cancelled. This creates a clear indicator for when a section is cancelled while retaining the original class status (e.g. open, full). 

## How Has This Been Tested?

The Websoc scraper was ran locally to accrue Winter 2026 data. 
SQL queries were then run locally.
Migration applied successfully to the local db.

API was run locally and Verified REST and GraphQL queries work on locally run API. 

```
anteater_api=# SELECT section_code, is_cancelled FROM websoc_section WHERE section_code = 34021 AND year = '2026' AND quarter = 'Winter';
 section_code | is_cancelled 
--------------+--------------
        34021 | t
```

REST:
```
> curl --silent "http://localhost:8787/v2/rest/websoc?year=2026&quarter=Winter&sectionCodes=34021" \
| jq ".data.schools[0].departments[0].courses[0].sections[0] | {sectionCode, isCancelled}"

{
  "sectionCode": "34021",
  "isCancelled": true
}
```

GraphQL:
```
> echo 'query WebsocTest {
  websoc(query: {
    year: "2026"
    quarter: Winter
    sectionCodes: "34021"
  }) {
    schools {
      departments {
        courses {
          sections {
            sectionCode
            isCancelled
            sectionComment
          }
        }
      }
    }
  }
}' \
| tr -d '\n' \
| jq --raw-input --raw-output --compact-output "{query: .}" \
| curl -s -X POST --data @- "http://localhost:8787/v2/graphql" -H "Content-Type: application/json" \
| jq ".data.websoc.schools[0].departments[0].courses[0].sections[0] | {sectionCode, isCancelled}"
{
  "sectionCode": "34021",
  "isCancelled": true
}
```

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/schema/websoc.ts
+++ b/apps/api/src/schema/websoc.ts
@@ -261,7 +261,7 @@ export const websocSectionSchema = z.object({
   }),
   isCancelled: z.boolean().openapi({
     description:
-      "Whether the section has been cancelled, independent of and superseding the status field. Imputed by searching the comment field of a section for a 'cancelled' indication.",
+      "Whether the section has been cancelled, independent of and superseding the status field. Imputed by searching the comment field of a section for a 'cancelled' or 'canceled' indication.",
   }),
   updatedAt: z.coerce.date(),
   webURL: z.string(),


iAmbrosial:
> Waiting for addition of resolution to #168.

Addition of resolution to #168 was added to the pr; ready for review.

apps/api/src/schema/websoc.ts
```diff
@@ -253,6 +253,11 @@ export const websocSectionSchema = z.object({
     totalEnrolled: z.string(),
     sectionEnrolled: z.string(),
   }),
+  isCancelled: z.boolean().openapi({
+    description:
+      "Explicitly indicates whether the section has been cancelled, superseding the 'status' property. Not present in original WebSoc data and is rather generated by parsing the section_comment column.",
```
laggycomputer: This is not very informative from the user's perspective:
* You should not mention a "column" because it's an implementation detail.
* "Parsing" the column is not sufficiently clear to gain any understanding about what this does.
* "Original data" introduces some complexity which doesn't benefit user comprehension.

Here are the elements which I think ought to appear in this description, some of which you've already included. The wording ought to be changed to be as concise as possible while being reasonably clear:
* A section can be "cancelled" per this field regardless of any other "status" which may be visible.
    This notion of cancellation supersedes any supposed "status".
* WebSoc does not actually have a formal notion of "cancellation".
* We instead infer cancellation by reading a comment which is placed manually on cancelled sections.
* This comment typically contains the word "cancelled".

For example, "Whether a section is cancelled, independent of the status field. Imputed by searching the comment field of the section for a 'cancelled' indication."

apps/api/src/schema/websoc.ts
```diff
@@ -253,6 +253,11 @@ export const websocSectionSchema = z.object({
...
+    example: false,
```
laggycomputer: Don't leave examples for very simple types like bools and most enums.

laggycomputer (changes requested):

sanskarm7 (approved):
tiny clarification, otherwise looks good - can merge right after

PR #271, opened 2025-11-29
deps: zod v4

laggycomputer:
migrate to zod v4 per https://zod.dev/v4/changelog

it typechecks properly i promise

## Related Issue

Closes #167

## Motivation and Context

zod v4 is faster and i like it

## How Has This Been Tested?

touched schemas/code run fine locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code involves a change to the database schema.
- [ ] My code requires a change to the documentation.

PR #278, opened 2025-12-06
refactor: addl zod v4 migration

laggycomputer:
he close my other pr #271 too early :(

Finishes that PR with additional stuff from the migration guide per https://github.com/icssc/anteater-api/pull/271#pullrequestreview-3547270034

### DIFF
--- a/apps/api/src/schema/ap-exams.ts
+++ b/apps/api/src/schema/ap-exams.ts
@@ -55,7 +55,7 @@ export const apExamsRewardSchema = z.object({
     description:
       "The number of units granted as generic elective credit (but not from any course) for this reward",
   }),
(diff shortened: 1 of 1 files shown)


sanskarm7 (changes requested):
mostly for sanity check, but i believe that [this](https://zod.dev/v4/changelog?id=updates-generics) may create some problems with ZodType as we've currently implemented.

may be misinterpreting though. for example:

```
export const programGroupRequirementSchema: z.ZodType<
  z.infer<typeof programRequirementBaseSchema> & {
    requirementType: "Group";
    requirementCount: number;
    requirements: z.infer<typeof programRequirementSchema>[];
  }
> = programRequirementBaseSchema
```
i believe zod now interprets input value as `unknown`, so we can no longer assume that it's the same at output. would this cause problems downstream?

if not, everything else looks good.

PR #281, opened 2025-12-07
deps: bump out CVEs and deprecations

laggycomputer:
CVEs are gone in Next (including the Big One with server components) and Hono.

Wrangler 3.100.0 was deprecated. Use 3.99.x instead.

Bump of `@graphql-tools/merge` so that it still works.


laggycomputer:
There was another pair of CVEs (lol). This isn't done yet.

sanskarm7:
ah, my bad. ok cool lmk how goes - no rush, the key manager is still broken anyways :(

PR #289, opened 2026-01-14
feat: add plaza verde study room scraper

ZackeryHe:
### Overview
We have expanded the Study Rooms endpoint to include study rooms (and study room slots) from Plaza Verde. This adds 17 new rooms, although the API treats each room as three separate rooms, one for each length of time the room can be reserved for (1 hour, 2 hours, and 3 hours), so 51 new study_room items have been added to our table. With this in mind, we now provide all available PV study room slots with a start time not before the current time, ending at end-of-day two days after the current day. Note that it only returns available slots, not busy ones, as we will discuss later.

### Our Strategy
To collect the information for both the study rooms and their slots, we followed these steps:
1. Query the Microsoft Bookings API that Plaza Verde uses to handle their study room reservations to collect all of the study rooms, including their name, ID, and other info.
2. Query it once more, this time using the collected study room IDs to request all of the study room availability for the current day and the following two days. PV doesn’t allow for reservations 3+ days in advance, so there would be no benefit to returning later slots, which would all technically be listed “available”.
3. Add the info for each study room to the study_room table in the database. This required some changes to the study_room schema, which we believe are necessary in order to add in the PV study rooms.
- We dropped the not_null constraints for techEnhanced and capacity, as those pieces of information aren’t provided by PV. All PV study rooms are listed with “null” in both of those columns.
-  We also added a new column called “url” for study_room, which gives a convenient link to the reservation page for a given study room. Currently, we only have URLs for the PV study rooms.
4. Parse the returned availabilities for each slot in order to present them in a usable way (their API returns the raw times a room is available, we transform that time into a list of available slots for each room).
- Note that start times are spaced out by 15 minutes despite being 1-3 hours long, which results in some of the slots for the same room overlapping each other.
- We also could have returned all of the 15 minute slots for each room (including availability/busyness), or each chunk of availability/busyness (i.e. when a room starts becoming available/busy to when that time ends). However, we believe that the way in which we present the information is optimal, as it matches the way in which PV itself presents it, and there is no clear benefit to returning busy times or 15/30 minute slots to match UCI Libraries’ style, as it would likely just cause confusion as to how the PV reservation system works.
5. Filter out study room slots that have already passed (i.e. those from earlier today) and add the info for each study room slot into the study_room_slot table.

### Schema Changes
We dropped the not_null constraints for techEnhanced and capacity, as those pieces of information aren’t provided by PV. All PV study rooms are listed with “null” in both of those columns.
We also added a new column called “url” for study_room, which gives a convenient link to the reservation page for a given study room. This only stores URLs for PV rooms: the API can already construct the URLs for UCI Libraries rooms, so those rows have a null in this column.

## Other Information

- The id for study locations in the past used to be numeric strings, e.g “44670” as they were taken directly from the UCI library system. The id for pv locations also taken from the booking system but instead look like: “fa1db88e-1cf9-4441-8e2f-0ae693add526”
- We no longer allow querying for hour windows that end before they start (e.g 03:00-01:00) as they resulted in unintended behavior for the entire study location api.

## How Has This Been Tested?
- normal run matches [online booking](https://outlook.office365.com/book/PlazaVerde3@americancampus.onmicrosoft.com/?ismsaljsauthenabled=true)
- rerunning properly clears old entries
- run during busy hour (such as early morning) matches
- run in NY timezone matches

## Types of changes

## Checklist:

- I dont know - My code requires a change to the documentation.


apps/data-pipeline/study-location-scraper/src/handler.ts
```diff
@@ -1,10 +1,11 @@
-import { doScrape } from "$lib";
+import { doScrape as doLibraryScrape } from "$lib";
 import { database } from "@packages/db";
+import { doScrape as doPVScrape } from "./pv";
 
 export default {
   async scheduled(_, env) {
     const db = database(env.DB.connectionString);
-    await doScrape(db);
+    await Promise.all([doLibraryScrape(db), doPVScrape(db)]);
```
laggycomputer: Keep in mind that `Promise.all` will **cancel** every promise still in flight if one promise rejects. This may lead to a Postgres connection hanging and a transaction staying open, for example. Either verify that these promises are correct to cancel or use [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled).

apps/data-pipeline/study-location-scraper/src/index.ts
```diff
@@ -1,12 +1,13 @@
 import { exit } from "node:process";
-import { doScrape } from "$lib";
+import { doScrape as doLibraryScrape } from "$lib";
 import { database } from "@packages/db";
+import { doScrape as doPVScrape } from "./pv";
```
laggycomputer: Import renaming is not good practice. Rename the original functions to these aliased names.

apps/data-pipeline/study-location-scraper/src/pv.ts
```diff
@@ -0,0 +1,365 @@
...
+  // have a map of rooms : its availability items
+  const availabilityMap = new Map<string, AvailabilityItem[]>();
+  for (const availability of availabilities) {
+    availabilityMap.set(availability.staffId, availability.availabilityItems);
+  }
+
+  const rooms: Array<typeof studyRoom.$inferInsert> = [];
+  const slots: Slot[] = [];
+
+  for (const service of services) {
```
laggycomputer: Why is this not a `.map`?

laggycomputer (changes requested):
Good work.
I'm very pessimistic about the use of non-UTC `Date` getters throughout this PR, which introduce a massive area for failure of locale portability.
After that, this code needs to be tested and thought through rigorously for timezone correctness.

There's also a conflict over the migration number, which can be fixed.

apps/data-pipeline/study-location-scraper/src/pv.ts
```diff
@@ -0,0 +1,361 @@
...
+  const hours = hoursMatch ? Number.parseInt(hoursMatch[1], 10) : 0;
+  const minutes = minutesMatch ? Number.parseInt(minutesMatch[1], 10) : 0;
+
+  return hours * 60 + minutes;
+}
+
+// given an array of staffIds, fetch the schedule within the given date
+async function fetchStaffAvailability(
+  staffIds: string[],
+  startDateTime: { dateTime: string; timeZone: string },
```
laggycomputer: This inline type is identical to the output of the `z.infer` type function at `typeof bookingsDateTimeSchema` and should be replaced with that type function.

apps/data-pipeline/study-location-scraper/src/pv.ts
```diff
@@ -0,0 +1,361 @@
...
+  const data = servicesResponseSchema.parse(await res.json());
+  return data.service;
+}
+
+/**
+ * Adds days to a date string formatted by Intl (M/D/YYYY or MM/DD/YYYY).
+ * Uses Date.UTC as a naive date container for calendar math,
+ * then formats back with timeZone: 'UTC' to avoid timezone conversion.
+ */
+function addDaysToDatePST(intlDateString: string, daysToAdd: number): string {
```
laggycomputer: This doesn't actually depend on PST, so it should not be named that way.

apps/data-pipeline/study-location-scraper/src/pv.ts
```diff
@@ -0,0 +1,361 @@
...
+ */
+function formatDateForAPI(intlDateString: string, boundary: "start" | "end") {
+  const [month, day, year] = intlDateString.split("/").map(Number);
+  const mm = String(month).padStart(2, "0");
+  const dd = String(day).padStart(2, "0");
+
+  return {
+    // note: the start time is 00:00 because of a bug where if you start in the
+    // middle of some BUSY interval, the API incorrectly assumes it's AVAILABLE.
+    // The filtering of past study room slots is handled in scrapePlazaVerde()
```
laggycomputer: This comment is a little vague. What does "start in the middle" mean? If this comment is going to live for a while, it should be understandable with minimal context.

laggycomputer (changes requested):

laggycomputer (changes requested):
The new comment is much clearer. This is good. Can you clarify how the last sentence about filtering out past slots relates to the several preceding sentences? What about the situation you describe requires you to filter out past rooms?

laggycomputer (approved):
The logic in the comment is strong. It would help to have one more sentence near the end explaining that we are assuming midnight is not in the middle of any interval, so setting the start time to midnight is guaranteed to not trigger the bug.

Even without that, this is an exemplary code comment. Well done!

laggycomputer:
@sanskarm7 Make sure you test with differing system and Postgres time zones. Setting both to distinct zones besides LA and UTC should suffice.

apps/api/src/schema/study-rooms.ts
sanskarm7: shouldn't we add something here to indicate that there is a separate internal system for Plaza Verde as well?

should something indicating that PV's ids don't follow a similar structure like UCI libraries as well?

ex:

UCI library id: `168432`

PV id: `fa1db88e-1cf9-4441-8e2f-0ae693add526`

sanskarm7 (changes requested):

apps/api/src/schema/study-rooms.ts
```diff
@@ -48,7 +48,7 @@ export const studyRoomSchema = z.object({
     description: "Additional data about this room, specifically for directions to the room",
```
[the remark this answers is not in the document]
ZackeryHe: note: `directions` likely misread as `description`

apps/api/src/schema/study-rooms.ts
```diff
@@ -25,7 +25,9 @@ export const slotSchema = z.object({
 
 export const studyRoomSchema = z.object({
   id: z.string().openapi({
-    description: "The ID of this study room, internal to the UCI Libraries system",
+    description:
+      "The ID of this study room taken from the external booking system. " +
+      "UCI Libraries IDs are numeric (e.g., 44670), while PV IDs are UUIDs (e.g., fa1db88e-1cf9-4441-8e2f-0ae693add526).",
```
laggycomputer: I would prefer we not leave examples in descriptions. This is best expressed as two statements in the description and two examples.

apps/data-pipeline/study-location-scraper/src/index.ts
```diff
@@ -1,13 +1,20 @@
...
 import { database } from "@packages/db";
+import { doPVScrape } from "./pv";
 
 async function main() {
   const url = process.env.DB_URL;
   if (!url) throw new Error("DB_URL not found");
   const db = database(url);
-  await doScrape(db);
-  exit(0);
+  const results = await Promise.allSettled([doLibraryScrape(db), doPVScrape(db)]);
```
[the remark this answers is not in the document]
ZackeryHe: In defense of `Promise.allSettled` we printed the stack trace through `status.reason` so we weren't losing it. I think `Promise.allsettled` is cleaner

apps/api/src/services/study-rooms.ts
```diff
@@ -20,14 +20,17 @@ export class StudyRoomsService {
...
         slots:
           sql`ARRAY_REMOVE(COALESCE(ARRAY_AGG(CASE WHEN ${studyRoomSlot.studyRoomId} IS NULL THEN NULL
             ELSE JSONB_BUILD_OBJECT(
               'studyRoomId', ${studyRoomSlot.studyRoomId},
               'start', TO_JSON(${studyRoomSlot.start} AT TIME ZONE 'America/Los_Angeles'),
               'end', TO_JSON(${studyRoomSlot.end} AT TIME ZONE 'America/Los_Angeles'),
-              'url', 'https://spaces.lib.uci.edu/space/' || ${studyRoom.id} || '?date=' || TO_CHAR(${studyRoomSlot.start}::timestamptz at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS"Z"') || '#submit_times', 
+              'url', CASE
+                WHEN ${studyRoom.url} IS NOT NULL THEN ${studyRoom.url}
+                ELSE 'https://spaces.lib.uci.edu/space/' || ${studyRoom.id} || '?date=' || TO_CHAR(${studyRoomSlot.start}::timestamptz at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS"Z"') || '#submit_times'
```
laggycomputer: You could do this with `COALESCE`, but this is fine too.

sanskarm7 (approved):
lgtm! fantastic work guys!!

PR #290, opened 2026-01-25
deps(key-manager): fix peer constraints

laggycomputer:
oopsie

followup to #281


laggycomputer:
yeah i'll just get around to this but this draft is getting stale

PR #291, opened 2026-01-26
feat: UCI dining API surface

laggycomputer:
We add a new full ETL pipeline taking data from the:
* MyDiningHub API as vendored for UCI, see https://uci.mydininghub.com/en. This provides information on one-off special events.
* The Adobe Commerce vendored system which provides information on menu-related entities, such as dishes and menus.

Our system is a port of the existing code in [PeterPlate](https://en.wikipedia.org/wiki/Peter_Plate).
We draw strongly on (and, where sensible, identically copy) their data model.

## Data collection and persistence

Acquisition of new data follows on the corresponding entrypoint [here](https://github.com/icssc/PeterPlate/blob/033abb0bacf33948d2f733dd21c675a7122b0c0d/packages/api/src/server/weekly/index.ts#L143) and downstream functions. Our new scraper is largely a straight port of this code, with the following salient design changes. First, code changes:

* The distinction between weekly and daily updates is abolished. As it stands in PeterPlate, a more extensive update is done weekly, with what are expected to be only minor menu revisions which are checked daily. No evidence has been provided to show that the daily trigger is sufficiently exhaustive, nor that is it is harmful to run the weekly tasks more frequently. A single trigger is simpler, anyway.
* PeterPlate code extensively suffers from an corollary of the [N+1](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) problem. In many cases, a single database query corresponds to the upsert of a single row. This incurs an entire network round-trip for each entity, of which there are thousands. This is often done many times consecutively, and concurrently with `Promise.all` or similar. In our implementation, rows are properly accumulated (and deduplicated where necessary) then used in a bulk upsert call.
* PeterPlate chose a more fragmented code organization, where code is spread across many files and packages and divided over "type of task" e.g. parsing, schemas, database upserting, etc. This supports re-use but has proven difficult to read and port. The main points of confusion has been difficulty in tracing the call stack of an operation upwards ("where is this exported function used?") and downwards ("what does this function being called accomplish?"), because parts of the same file are imported by various tasks with disjoint subject matters ("why am I back in this file again?"). Our scraper takes a new approach where files are imported hierarchically. Significantly sized files import smaller primitives and export only a single entrypoint and types describing passed and returned values. They hide their implementation details, such as GraphQL query strings (which are currently exported in PeterPlate), so that there is no need to reason or speculate about unexpected usage of symbols elsewhere. The ideal is that every file is only imported in one or two other places. Small items used in exactly one place have mostly been inlined. The new scraper files form a clear implicit hierarchy, where the overall entrypoint orchestrates several of these complex functions in sequence to perform work:
  * `doScrape` in `lib.ts` invokes the flagship functions exported from `update-events.ts` and `update-restaurants.ts`.
  * These two functions draw on private code in the same file and the `fetch-*.ts` files, which each perform querying and parsing tightly related to a single API query.
  * All files mentioned so far may draw from small utility files, which define sufficiently trivial functions and types.
* We remove the test suite. Maybe we'll add it back later :(
* PeterPlate currently uses scheduled AWS Lambda functions to invoke their scraping code. We port to Cloudflare Workers, the home of API, but this is as simple as changing the root entrypoint.
* We remove `winston`. A full logging system is not possible in the Cloudflare Workers context, because we have no disk access. As is the norm with other API scrapers which run in the cloud, we use the built-in `console` API to use Cloudflare Workers observability as a logging solution. To compensate:
  * Logs have been made more verbose about their surrounding context.
  * Use of `Promise.all` and `Promise.allSettled` is reduced, since they often prevent exception backtracing from reaching relevant frames in our code. The entire scraping process takes seconds at most, so there is no need to call upon more concurrency.
* We remove `axios`. This dependency is not necessary to achieve our relatively simple network requests.
* We migrate to Zod v4.
* Various code style nits have been enforced.

Notable database changes:
* We remove tracking of a "created at" timestamp for entities. It is not clear that this is necessary.
* Table names are changed to use singular number to conform to existing API conventions. Dining-related tables have an appropriate name prefix, because there are many such tables and their meaning may not be clear out of context. This has precedent with WebSoc related tables *e.g.* `websoc_section`.
* PeterPlate used Drizzle triggers to automatically set `updatedAt` to a `new Date()` (*i.e.* the time at that moment) for insert. However, it is not necessarily true that the data is/was correct at database insert time. The latest correct time is is the last timestamp we can observe before yielding execution control for the network request which acquires the data. We collect this time and store it into the database as appropriate, manually setting `updatedAt` on inserted entities. Drizzle triggers are not used. We will suffer from slight timing imprecision due to the [Cloudflare Workers security model](https://developers.cloudflare.com/workers/reference/security-model/), but this is preferable to being inaccurate.
* The column on dish entities which stores a menu ID is redundant because a many-to-many join table provides strictly more functionality. The restaurant ID and name enums are removed, since they are unnecessarily restrictive. See https://github.com/icssc/PeterPlate/issues/602.
* Menu entities are removed. Menu IDs were generated on-the-fly from a restaurant ID, date, and period number anyway, so there were always one or zero menus per period. This is an excess layer of indirection, so dishes now directly belong to a period.
* Periods now have a UUID primary key with a unique index on (restaurant ID, date, period number).
* PeterPlate uses the Drizzle relations v1 API ([docs](https://orm.drizzle.team/docs/relations)). We omit this because it is not normative under API's usage of Drizzle primarily as a SQL wrapper.
* Data not relevant to a general-purpose web API, *e.g.* ratings data on dishes, is no longer stored. This sort of data remains PeterPlate's concern and will stay behind.
* Various column defaults have been removed in favor of nullability. It has been decided that the underlying missing state should be conveyed to clients unmodified. It is a client concern, not ours, what to do with (such as how to render) missing data and we shall not fill it in on their behalf.

Still to do:
- [x] Add more indexes, pending me seeing database query code which I have not examined closely.
- [x] We should definitely be using UUID surrogate keys more.
- [ ] ~~Use codegen of some kind to derive GraphQL types for queries instead of manually defining schemas and types~~ Nah.
- [x] Change nutrition scalars to decimals and return them as `Number` as requested by PeterPlate
- [x] Add image URLs for dishes https://github.com/icssc/PeterPlate/pull/621
- [x] Drop old restaurant ID (numbers), these are hardcoded backward compat we don't need anymore

## API layer
* Events, Dish, Dates routes are closely based after the existing PeterPlate tRPC routes
* Events route returns all dining events that end today or later. Dish route takes an ID and returns a single dish with nutrition and dietary restriction information. Dates route returns the earliest and latest dates that have menu information available.
* Restaurants and RestaurantToday routes are adapted from the PeterPlate Zotmeal/Peterplate route. The original implementation returned all restaurant data that PeterPlate used for their app in one response, which is not ideal for an API to serve an app with unique needs. It should be more flexible, which is why this has been refactored into three separate routes: Restaurants, Batch Dishes, and Restaurant Today. The reason for this split is that restaurants and stations don't change much and are likely to support UI elements. Restaurant Today returns the restaurant information for a single day, and this limits data size by requiring queries by day and restaurant ID. Dishes are likely to repeat across multiple menus, so we can make the data more concise by not repeating their data every time their Dish ID appears. 
* GraphQL endpoints were implemented in parallel with REST routes to provide equivalent functionality (Events, Dish, Dates, Restaurants, RestaurantToday)

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/rest/routes/dining.ts
+++ b/apps/api/src/rest/routes/dining.ts
@@ -23,7 +23,7 @@ const diningRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
 diningRouter.get("*", productionCache({ cacheName: "anteater-api", cacheControl: "max-age=3600" }));
 
 const eventsRoute = createRoute({
-  summary: "Get upcoming dining events",
+  summary: "Get dining events",
   operationId: "getDiningEvents",
   tags: ["Dining"],
   method: "get",


apps/data-pipeline/dining-scraper/src/update-restaurant.ts
```diff
@@ -73,20 +73,22 @@ export async function updateRestaurant(
...
-  const daysUntilNextSunday = (7 - todayDayOfWeek) % 7;
-  for (let i = 0; i <= daysUntilNextSunday; i++) {
+  const daysUntilNextSunday = (7 - todayDayOfWeek) % 7 || 7;
+  for (let i = 0; i <= daysUntilNextSunday; ++i) {
     const nextDate = new Date(today);
     nextDate.setDate(today.getDate() + i);
     datesToFetch.push(nextDate);
   }
+  console.log(
+    `Fetching ${datesToFetch.length} dates: ${datesToFetch.map((d) => format(d, "yyyy-MM-dd")).join(", ")}`,
```
laggycomputer: we are not fetching anything here, so this nomenclature is not correct

laggycomputer:
We could, but these are legacy JSDocs from PeterPlate and we don't write these ourselves in most of our code, so I don't feel particularly compelled to expand them. The types are mostly enough.

apps/api/src/schema/dining.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const eventSchema = z.object({
+  title: z.string().openapi({ example: "Lunar New Year Celebration" }),
+  image: z.string().nullable().openapi({ example: "https://images.elevate-dxp.com/..." }),
+  restaurantId: z.string().openapi({ example: "3056" }),
+  longDescription: z.string().nullable(),
+  // transforms Date objects in db to ISO 8601 strings (I hope)
+  start: z.coerce
+    .date()
+    .nullable()
```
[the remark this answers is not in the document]
iAmbrosial: key key key

apps/api/src/services/dining.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (input.restaurantId) {
+      conds.push(eq(diningEvent.restaurantId, input.restaurantId));
+    }
+
+    return this.db
+      .select({
+        title: diningEvent.title,
+        image: diningEvent.image,
+        restaurantId: diningEvent.restaurantId,
+        // somewhere it was mentioned that shortDescription is never stored, so don't return it
```
iAmbrosial: For some reason I recall this being said somewhere and looking at the scraped data backs this claim up, but I can't be 100% certain that it's right – though the choice is probably fine for now.

apps/api/src/rest/routes/dining.ts
```diff
@@ -0,0 +1,263 @@
...
+      content: { "application/json": { schema: errorSchema } },
+      description: "Server error occurred",
+    },
+  },
+});
+
+const batchDishesRoute = createRoute({
+  summary: "Retrieve dishes by IDs",
+  operationId: "batchDishes",
+  tags: ["Courses"],
```
sanskarm7: marked as the wrong tag - won't show under dishes

apps/api/src/rest/routes/dining.ts
```diff
@@ -0,0 +1,263 @@
...
+  );
+});
+
+diningRouter.openapi(dishRoute, async (c) => {
+  const { id } = c.req.valid("param");
+  const service = new DiningService(database(c.env.DB.connectionString));
+  const dish = await service.getDishById({ id });
+
+  if (!dish) {
+    return c.json({ ok: false, message: "Dish not found" }, 404);
```
sanskarm7: this doesn't ever actually give you this message

am consistently getting : `Invalid input: expected string, received undefined`

please double check case for other endpoints as well

laggycomputer (commented):
`@cacheControl` should be on all graphql schemas iirc

apps/api/src/services/dining.ts
```diff
@@ -0,0 +1,272 @@
...
+
+type DiningDishQuery = z.infer<typeof dishQuerySchema>;
+type DiningEventQuery = z.infer<typeof diningEventsQuerySchema>;
+
+export class DiningService {
+  constructor(private readonly db: ReturnType<typeof database>) {}
+
+  async getUpcomingEvents(input: DiningEventQuery) {
+    // only get events ending at or after the current time
+    const conds = [gte(diningEvent.end, sql`NOW()`)];
```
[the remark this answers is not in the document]
iAmbrosial: added functionality to return events ending at or after the current time, or events with a null (indefinite) end date

apps/api/src/rest/routes/dining.ts
```diff
@@ -0,0 +1,263 @@
...
+
+diningRouter.openapi(eventsRoute, async (c) => {
+  const query = c.req.valid("query");
+  const service = new DiningService(database(c.env.DB.connectionString));
+  const events = await service.getUpcomingEvents(query);
+
+  return c.json({ ok: true, data: diningEventsResponseSchema.parse(events) }, 200);
+});
+
+diningRouter.openapi(batchDishesRoute, async (c) => {
```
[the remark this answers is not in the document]
iAmbrosial: this could probably go either way tbh

apps/api/src/rest/routes/dining.ts
iAmbrosial: sorry, when do you get this msg? on my end if you provide a non-existent dish ID that message is returned for the dish route – though the batch dishes route just returns an empty array instead

apps/api/src/graphql/schema/dining.ts
```diff
@@ -0,0 +1,127 @@
...
+  date: String
+}
+
+extend type Query {
+  diningEvents(query: DiningEventsQuery): [DiningEvent!]!
+  diningDish(id: String!): DiningDish
+  batchDiningDishes(ids: [String!]!): [DiningDish!]!
+  diningDates: DiningDates!
+  getRestaurants(query: RestaurantsQuerySchema): [Restaurant!]!
+  getRestaurantToday(query: RestaurantTodayQuerySchema): RestaurantToday
```
laggycomputer: Any query with mandatory arguments should be non-nullable.

PR #295, opened 2026-01-29
feat(degreeworks-scraper): Label Optional Specializations

### LINKED ISSUE #219: Determine which specializations are optional
#218 provides optional specializations, but we still don't know which are actually optional.
Not having this information isn't too harmful, since the specialization block could simply be ignored or not fetched, but it is technically incorrect.

### LINKED ISSUE #244: Docs claim specializations are mandatory
Related to #219.

The docs currently claim:
https://github.com/icssc/anteater-api/blob/c2512458f9278d4f307aafcee944077c735c1f0a/apps/api/src/rest/routes/programs.ts#L116

But that's clearly not true. For example, [every specialization of Chemistry, B.S. is optional](https://catalogue.uci.edu/schoolofphysicalsciences/departmentofchemistry/chemistry_bs/#requirementstext). We should fix the docs accordingly.

HwijungK:
## Description

When a call to DW API for a major is made without a corresponding specialization and that major requires a spec, DW will return a rule that indicates that a specialization is required. This rule is almost always a `Course` rule (the odd exception being B.A. Classics which gives a`Block` rule), has the property `IfElsePart:ElsePart`, and has a `proxyAdvice` with `textList` containing any of following words: specialization, concentration, emphasis, track, area, major.

This Pattern is used to create a new DegreeWorksRequirement subtype, DegreeWorksSpecRequirement, which denotes that a major requires a specialization to complete.

**Note**: For many graduate programs, especially PHDs, degreeworks doesn't officially have a specialization. For example,
MS in Civil and Environmental Engineering claims to have the specializations:
(a) Structures, Geotechnics, and Materials;","(b) Transportation Systems; (c) Hydrology and Water Resources Systems; (d) Environmental and Energy Systems. However, none of these specializations are found to exist.
For such cases, `specializationRequired` is set to false after matching majors to spec.

## Related Issue

Closes #219 
Also closes #244 which is a first-task, but it would be silly not to change it on this occasion.

## Motivation and Context

Look at issue Above.

## How Has This Been Tested?

A Selection of degrees (labeled OK) were manually checked against UCI course catalogue to check for correctness.

The following degrees will be flagged to have requiredSpecializations (collectively 50 of the 108 specializations)
"MFA in Drama" (majorCode = 231, degree = MFA)  ------------------- **OK**
"MFA in Music" (majorCode = 579, degree = MFA)
"Major in Music" (majorCode = 579, degree = BA)  -------------------- **OK** 
"Major in Music Performance" (majorCode = 582, degree = BMUS)
"Major in Classics" (majorCode = 174, degree = BA)                             -------  **OK**
"Major in Electrical Engineering" (majorCode = 276, degree = BS)     -------     **OK**
"Major in Civil Engineering" (majorCode = 284, degree = BS)        -------           **OK**
"Major in Chemical Engineering" (majorCode = 294, degree = BS)    -------     **Hard coded exception (see below)**
"Major in Computer Science" (majorCode = 201, degree = BS)       -------        **OK**
"Major in Chinese Studies" (majorCode = 163, degree = BA)   ------------- **OK**
"Ph.D. in Education" (majorCode = 0B0, degree = PHD)
"Master of Arts in Teaching" (majorCode = 25B, degree = MAT)
"MS in Pharmacological Sciences" (majorCode = 07T, degree = MS)

The following degrees will be flagged as having optional specializations
"Major in European Studies" (majorCode = 762, degree = BA)                       --------------                **OK**
"Major in Environmental Science and Policy" (majorCode = 0DS, degree = BA).    -------     **OK**
"Major in Earth System Science" (majorCode = 763, degree = BS)         -------                      **OK**
"Major in Chemistry" (majorCode = 153, degree = BS)               -------                                    **OK**
"Major in Philosophy" (majorCode = 651, degree = BA)             -------                                    **OK** -- Polymorphic Major
"M.S. in Physics" (majorCode = 666, degree = MS)                   -------          (doesn't have a catalogue page, but DW looks OK)
"Major in English" (majorCode = 345, degree = BA)                   -------              **OK? (this one weird, check screenshots)
"Major in Materials Science and Engineering" (majorCode = 328, degree = BS).      **OK**
"Major in Biomedical Engineering" (majorCode = 289, degree = BS)             -------              **OK**
"Major in History" (majorCode = 429, degree = BA)                                -------                       **same as BA English**
"M.S. in Mathematics" (majorCode = 540, degree = MS)                        -------                      **OK**
"Major in Physics" (majorCode = 666, degree = BS)                               -------                       **OK**
"Major in Mechanical Engineering" (majorCode = 277, degree = BS)     -------                      **OK**
"Major in Mathematics" (majorCode = 540, degree = BS)
"Major in Biomedical Engineering: Premedical" (majorCode = 28B, degree = BS)
"Major in Applied Physics" (majorCode = 075, degree = BS)
"Major in Comparative Literature" (majorCode = 192, degree = BA)
"Major in Applied and Computational Mathematics" (majorCode = 0K6, degree = BS)  -------   **OK**

NOTE:
chemical engineering: the textList states "16 units of approved technical electives or contact advisor to select a specialization," so it's an optional spec but it gets flagged.

Ending Notes:

image

image

generally, a spec is refered to from a major block through a block rule w/ requirement type SPEC
A English major without the teach specialization returns the actual requirements in a "OTHER" block with the major block existing only to point to said "OTHER" block. example of polymorphic block, but might be hinting at something much weirder.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/rest/routes/ap-exams.ts
+++ b/apps/api/src/rest/routes/ap-exams.ts
@@ -4,8 +4,9 @@ import {
   apExamsQuerySchema,
   apExamsResponseSchema,
   coursesGrantedTreeSchema,
-  errorSchema,
-  responseSchema,
+  response200,
+  response404,
+  response500,
 } from "$schema";
 import { APExamsService } from "$services";
 import { OpenAPIHono, createRoute } from "@hono/zod-openapi";
@@ -25,18 +26,9 @@ const apExamsRoute = createRoute({
     "Get AP exam data: mappings from College Board exam names to UCI Catalogue names and course credit for each exam score.",
   request: { query: apExamsQuerySchema },
   responses: {
-    200: {
-      content: { "application/json": { schema: responseSchema(apExamsResponseSchema) } },
-      description: "Successful operation",
-    },
-    404: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "AP Exam mapping not found",
-    },
-    500: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Server error occurred",
-    },
+    200: response200(apExamsResponseSchema),
+    404: response404("AP Exam mapping not found"),
+    500: response500(),
   },
 });
 
--- a/apps/api/src/rest/routes/calendar.ts
+++ b/apps/api/src/rest/routes/calendar.ts
@@ -1,6 +1,13 @@
 import { defaultHook } from "$hooks";
 import { productionCache } from "$middleware";
-import { calendarQuerySchema, calendarTermSchema, errorSchema, responseSchema } from "$schema";
+import {
+  calendarQuerySchema,
+  calendarTermSchema,
+  response200,
+  response404,
+  response422,
+  response500,
+} from "$schema";
 import { CalendarService } from "$services";
 import { OpenAPIHono, createRoute } from "@hono/zod-openapi";
 import { database } from "@packages/db";
@@ -16,24 +23,10 @@ const calendarTermRoute = createRoute({
   request: { query: calendarQuerySchema },
   description: "Retrieves key dates for the provided term.",
   responses: {
-    200: {
-      content: {
-        "application/json": { schema: responseSchema(calendarTermSchema) },
-      },
-      description: "Successful operation",
-    },
-    404: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Term not found",
-    },
-    422: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Parameters failed validation",
-    },
-    500: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Server error occurred",
-    },
+    200: response200(calendarTermSchema),
+    404: response404("Term not found"),
+    422: response422(),
+    500: response500(),
   },
 });
 
@@ -45,18 +38,8 @@ const allCalendarTermsRoute = createRoute({
   path: "/all",
   description: "Retrieves all data for all terms that are currently available.",
   responses: {
-    200: {
-      content: {
-        "application/json": {
-          schema: responseSchema(calendarTermSchema.array()),
-        },
-      },
-      description: "Successful operation",
-    },
-    500: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Server error occurred",
-    },
+    200: response200(calendarTermSchema.array()),
+    500: response500(),
   },
 });
 
--- a/apps/api/src/rest/routes/catalogue.ts
+++ b/apps/api/src/rest/routes/catalogue.ts
@@ -1,8 +1,9 @@
 import { defaultHook } from "$hooks";
 import { productionCache } from "$middleware";
 import {
-  errorSchema,
-  responseSchema,
+  response200,
+  response404,
+  response500,
   sampleProgramsQuerySchema,
   sampleProgramsResponseSchema,
 } from "$schema";
@@ -21,22 +22,9 @@ const sampleProgramsRoute = createRoute({
   description: "List sample programs in UCI's current catalogue.",
   request: { query: sampleProgramsQuerySchema },
   responses: {
-    200: {
-      content: {
-        "application/json": {
-          schema: responseSchema(sampleProgramsResponseSchema),
-        },
-      },
-      description: "Successful operation",
-    },
-    404: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Sample program data not found",
-    },
-    500: {
-      content: { "application/json": { schema: errorSchema } },
-      description: "Server error occurred",
-    },
+    200: response200(sampleProgramsResponseSchema),
+    404: response404("Sample program data not found"),
+    500: response500(),
   },
 });
 
--- a/apps/api/src/rest/routes/courses.ts
+++ b/apps/api/src/rest/routes/courses.ts
@@ -6,11 +6,12 @@ import {
   coursesByCursorQuerySchema,
   coursesPathSchema,
   coursesQuerySchema,
-  cursorResponseSchema,
-  errorSchema,
   prerequisiteSchema,
   prerequisiteTreeSchema,
-  responseSchema,
+  response200,
+  response404,
+  response422,
+  response500,
 } from "$schema";
 import { CoursesService } from "$services";
 import { OpenAPIHono, createRoute } from "@hono/zod-openapi";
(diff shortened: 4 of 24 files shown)


laggycomputer (changes requested):
Name your schema migration.

package.json
```diff
@@ -28,7 +28,7 @@
     "tsx": "4.19.2",
     "typescript": "5.6.3"
   },
-  "packageManager": "pnpm@10.17.0",
+  "packageManager": "pnpm@10.28.2",
```
laggycomputer: Was this a mistake? We shouldn't need to touch dependencies

packages/db/src/schema.ts
```diff
@@ -128,6 +128,10 @@ export type DegreeWorksMarkerRequirement = {
   requirementType: "Marker";
 };
 
+export type DegreeWorksSpecRequirement = {
+  requirementType: "Spec";
+};
```
laggycomputer: We need to think carefully about this.
- PeterPortal handles each requirement type separately, so adding this will require custom handling.
- We can mitigate the above by making this new type a sub-type of marker requirements, in some way which is detectable by inspecting the object structure.
- Maybe we should remove this entirely, since it is redundant with the new table column, and this isn't really a real requirement because its semantics do not make sense without reference to its context (unlike other requirements which can be meaningful at any position in the tree).

Let's discuss today.

andrew-wang0:
Expanding on @laggycomputer comments. You'll want to undo your two migrations then run a fresh migration setting the `--name` flag for a descriptive name (drizzle automatically assigns funky names if you don't set this)

See docs here: https://orm.drizzle.team/docs/drizzle-kit-migrate

package.json
andrew-wang0: We should look into updating this when we get the chance though!

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -112,93 +112,105 @@ export class AuditParser {
...
-            await Promise.all(includedCourses.map(this.normalizeCourseId.bind(this))).then((x) =>
-              x.flat().map((y) => [y.id, y]),
-            ),
-          );
-          const excludedCourses =
-            rule.requirement.except?.courseArray.map(
+      if (
+        rule.ifElsePart === "ElsePart" &&
+        rule.proxyAdvice?.textList.some((x) =>
+          /specialization|concentration|emphasis|area|track|major/i.test(x),
```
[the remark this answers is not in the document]
laggycomputer: No, they are not only this word because they are a human-readable statement specifying that the student should pick a specialization e.g. "Select a specialization".

packages/db/src/schema.ts
```diff
@@ -671,6 +677,7 @@ export const major = pgTable(
       .notNull(),
     code: varchar("code").notNull(),
     name: varchar("name").notNull(),
+    requireSpec: boolean("require_spec").notNull(),
```
laggycomputer: Can we name this `specializationRequired`?

packages/db/src/schema.ts
laggycomputer: > My tentative plan for it is to create a RuleGroup for polymorphic majors that requires a user to either choose a spec or choose additional courses that are required for students who don't specialize.

I would *strongly* prefer to not attempt to stitch together requirements in this way. For example, a major may have additional courses but only with certain specializations present. That plan is too inflexible.

> we could choose to view the 'pure' form of a polymorphic major as a specialization of it's own

Not unless that major does not require a specialization. In which case, making a new "specialization" is still less correct than just allowing null spec.

> Can you explain how making spec requirement a sub-type of marker requirement would help.

All data returned needs to meaningful for clients, not just for our implementation details. If you do not make this new sub-type of `Requirement` a sub-type of `MarkerRequirement`, it will require additional parsing code from clients.

> how does PP handle regular markers as of now.?

They get rendered as check-boxes labelled with the title of the requirement.

> I need some way to mark inside the AuditParser that a program requires a specialization, and process that later on in index.ts, so it feels like some kind of requirement is necessary.

This new requirement is still not the best way to achieve that for several reasons, including difficulty of retrieval. Create or extend a tuple or aggregate containing program data to include whether a specialization is required.

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -406,6 +406,11 @@ export class Scraper {
       this.parsedPrograms.set("Major in Art History", x);
     }
 
+    // Chemical Engineering is falsely marked as requiring a specialization (pr#295)
+
+    const chemE = this.parsedPrograms.get("Major in Chemical Engineering") as MajorProgram;
```
laggycomputer: This may fail. Check your access, the way it is done for Art History elsewhere.

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -306,7 +306,14 @@ export class Scraper {
 
         if (got !== null) {
           specBlock = got.block;
-          foundMajor = got.parent;
+          const majorProgram = this.parsedPrograms.get(got.parent.name);
+          if (majorProgram) {
+            foundMajor = majorProgram[1];
+          } else {
+            console.log(
+              `warning: ${specName} (spec code = ${specCode}) is related to ${got.parent.name} in spec cache but no such major exists`,
```
laggycomputer: Could be shorter:

warning: (spec) has cached association to (nonexistent/not found/etc) major (code)

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -400,12 +413,21 @@ export class Scraper {
...
+      x[1].specializationRequired = false;
       x[1].requirements = [...x[1].requirements, ...y.requirements, ...z.requirements];
       this.parsedSpecializations.delete("AHGEO");
       this.parsedSpecializations.delete("AHPER");
       this.parsedPrograms.set("Major in Art History", x);
     }
 
+    // Chemical Engineering is falsely marked as requiring a specialization (pr#295)
+
+    const chemE = this.parsedPrograms.get("Major in Chemical Engineering") as MajorProgram;
```
laggycomputer: This cast isn't correct and I think it's likely no cast is required, because the map value type is known and you are inside a truthy guard when attempting to access.

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -306,7 +306,14 @@ export class Scraper {
 
         if (got !== null) {
           specBlock = got.block;
-          foundMajor = got.parent;
+          const majorProgram = this.parsedPrograms.get(got.parent.name);
+          if (majorProgram) {
+            foundMajor = majorProgram[1];
+          } else {
+            console.log(
+              `warning: ${specName} has cached relation to non-existant major, ${got.parent.name} (spec code = ${specCode})`,
```
laggycomputer: It's spelled existent...

PR #300, opened 2026-02-05
feat(websoc-scraper): Dynamic Resolution for Enrollment History

### LINKED ISSUE #134: RFC: Higher Resolution for Enrollment History
Currently, we store enrollment data on a per day basis. While is likely sufficient for most use cases, it would be nice to have higher resolution data during active enrollment.

One benefit I can think of is allowing users to better gauge if they'll be able to enroll (though this is really only important if your enrollment period is the day that the class fills up, and even then it's questionably useful). This would allow us to show the user the average number enrolled during their enrollment period on AA, which could be a really useful feature that differentiates AA.

* Is this worth the extra endpoint / complexity?
  * This, however, is lower than I initially expected since the scraper already runs every 5 min. So it would just be a matter of checking the calendar for the dates.
* How often should the increased period be?
* How should the data be returned to the user?
  * Our current time series returns dates without timestamps. While we could change modify the endpoint to include times with a resolution param, it seems neater (especially for types) to just create a new endpoint.

cc: @MinhxNguyen7, @andrew-wang0 
I see that you and Douglas authored the Prior Enrollment panel on AA (https://github.com/icssc/AntAlmanac/commit/210298dbc20a1b0f937b5da1207afdfa77458acc), so any input on how to best format the data would be great.

aadi-shanker:
## Description

Implements variable-frequency enrollment snapshots based on academic calendar periods. The scraper now captures enrollment data at different frequencies depending on enrollment activity:

ENROLLMENT period (Week 8-10): Every 3 hours, 7am-7pm only (~5 snapshots/day)
ADD_DROP period (Week 1-2): Every 6 hours 24/7, with hourly snapshots on Week 2 Friday 12pm-5pm (~4-8 snapshots/day)
REGULAR period (Week 3-7): Once per week (168 hours)
Between quarters: Once per week for late enrollment

Database-driven frequency tracking checks hours elapsed since the last snapshot and only inserts when thresholds are met. This prevents missed snapshots during scraper outages (self-correcting behavior) while optimizing storage usage.
## Related Issue

#134

## Motivation and Context

Missing intraday enrollment trends -Want to increase and visualize hourly changes during critical periods

## How Has This Been Tested?

I ran a test file with multiple detection unit tests based on given fake dates. I also ran the scraper using fake dates just to double check the processing logic is correct.

## Types of changes

## Checklist:


laggycomputer (changes requested):
As we're discussing right now:
- Don't duplicate week logic, move what you need into stdlib
- Make sure week still works
- `shouldSnapshotThisRun` should be computed in the top-level scrape function, then passed into `doChunkUpsert`, don't use a module global
- Make sure enrollment history endpoints still work
- More code comments; what does your logic *mean*?

ParzivalPerhaps (approved):
Aadi owes me $10,000

lgtm

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -810,6 +841,82 @@ export async function scrapeTerm(
...
+    console.log(
+      `Period: ${period}, Week: ${week}, Day: ${period === "ADD_DROP" ? addDropDay : "N/A"}, ` +
+        `Hours since last: ${hoursSinceLastSnapshot.toFixed(1)}, Should snapshot: ${shouldSnapshotThisRun}`,
+    );
+  } else {
+    // Between quarters: snapshot weekly
+    shouldSnapshotThisRun = hoursSinceLastSnapshot >= 168;
+    console.log(
+      `Between quarters, Hours since last: ${hoursSinceLastSnapshot.toFixed(1)}, ` +
+        `Should snapshot: ${shouldSnapshotThisRun}`,
```
laggycomputer: "Should" should not be capitalized throughout

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -50,6 +52,27 @@ const SECTIONS_PER_CHUNK = 891;
  */
 const LAST_SECTION_CODE = "97999";
 
+/**
+ * Snapshot decision calculated once per scraper run.
+ * Ensures all chunks use identical timestamp to prevent partial snapshots.
+ */
+type SnapshotDecision = {
+  shouldSnapshot: boolean;
+  timestamp: Date;
```
laggycomputer: What timestamp is this? Is this when we need to scrape? Is this when the last scrape happened? Rename and possibly document.

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -810,6 +841,82 @@ export async function scrapeTerm(
...
+    }
+
+    const addDropDay =
+      Math.floor(
+        (scraperRunTimestamp.getTime() - currentTerm.instructionStart.getTime()) / DAY_MS,
+      ) + 1;
+    console.log(
+      `Period: ${period}, Week: ${week}, Day: ${period === "ADD_DROP" ? addDropDay : "N/A"}, ` +
+        `Hours since last: ${hoursSinceLastSnapshot.toFixed(1)}, Should snapshot: ${shouldSnapshotThisRun}`,
+    );
```
laggycomputer: (Acting) as an uninformed reader, I don't understand what `addDropDay` is. I would assume it's the day on which the add drop deadline for this quarter falls, but why do we need it? Why do we just print it out? If I were looking at the logs, I'd similarly be confused.

laggycomputer (changes requested):

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -810,6 +862,76 @@ export async function scrapeTerm(
...
+        lte(calendarTerm.instructionStart, scraperRunTimestamp),
+        gte(calendarTerm.instructionEnd, scraperRunTimestamp),
+      ),
+    )
+    .limit(1);
+
+  // Get last snapshot to calculate time elapsed
+  const [lastSnapshot] = await db
+    .select({
+      createdAt: sql<string>`MAX(${websocSectionEnrollment.createdAt})`.as("created_at"),
```
laggycomputer: Use builtin: https://orm.drizzle.team/docs/select#aggregations-helpers

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -810,6 +862,76 @@ export async function scrapeTerm(
...
+    const period = detectPeriod(week);
+    const { hour, currentDayOfWeek } = getPacificTimeParts(scraperRunTimestamp);
+
+    switch (period) {
+      case "ENROLLMENT":
+        // ENROLLMENT (weeks 8-10): snapshot every 3 hours during 7am-7pm
+        shouldSnapshotThisRun = hour >= 7 && hour < 19 && hoursSinceLastSnapshot >= 3;
+        break;
+      case "ADD_DROP":
+        // Week 2 Friday 12pm-5pm: hourly snapshots (add/drop deadline)
```
laggycomputer: This comment is a series of somewhat disorganized points. You mean to say this:

- The add/drop deadline is Friday of week 2 at 5 pm.
- We expect a lot of activity between noon and 5.
- We will snapshot hourly during this time

Try to rewrite this comment to provide a more concise chain of thought.

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -810,6 +862,76 @@ export async function scrapeTerm(
...
+        if (week === 2 && currentDayOfWeek === 5 && hour >= 12 && hour <= 17) {
+          shouldSnapshotThisRun = hoursSinceLastSnapshot >= 1;
+        } else {
+          // ADD_DROP (weeks 0-2): snapshot every 6 hours
```
laggycomputer: No need to repeat ADD_DROP here

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -50,6 +52,52 @@ const SECTIONS_PER_CHUNK = 891;
...
+type SnapshotDecision = {
+  shouldSnapshot: boolean;
+  scraperRunStartTime: Date;
+};
+
+type EnrollmentSnapshotPeriod = "ADD_DROP" | "ENROLLMENT" | "REGULAR";
+
+/**
+ * Determine the academic period based on week number.
+ * Week 0-2: ADD_DROP, Week 8-10: ENROLLMENT, Other: REGULAR
```
laggycomputer: This line is trivially inferrable from the simple implementation and should be removed.

laggycomputer (changes requested):

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -50,6 +52,49 @@ const SECTIONS_PER_CHUNK = 891;
  */
 const LAST_SECTION_CODE = "97999";
 
+/**
+ * Decision on whether to snapshot enrollment data during current scraper run.
+ * Calculated once before processing chunks based on academic period and time elapsed from last snapshot.
```
laggycomputer: Remove "academic" here too

laggycomputer (changes requested):

laggycomputer (changes requested):

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -810,6 +859,76 @@ export async function scrapeTerm(
...
+      and(
+        lte(calendarTerm.instructionStart, scraperTimestamp),
+        gte(calendarTerm.instructionEnd, scraperTimestamp),
+      ),
+    )
+    .limit(1);
+
+  // Get last snapshot to calculate time elapsed
+  const lastSnapshot = await db
+    .select({ createdAt: max(websocSectionEnrollment.createdAt) })
```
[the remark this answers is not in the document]
aadi-shanker: I run into an issue when i move max out of an object like this .select(max(websocSectionEnrollment.createdAt))
I am getting hit with Argument of type 'SQL<Date | null>' is not assignable to parameter of type 'SelectedFields'. Index signature for type 'string' is missing in type 'SQL<Date | null>'.

packages/db/src/schema.ts
```diff
@@ -179,9 +180,10 @@ export const sampleProgramVariation = pgTable("sample_program_variation", {
 
 // Misc. enums
 
-export const terms = ["Fall", "Winter", "Spring", "Summer1", "Summer10wk", "Summer2"] as const;
+export { terms } from "@packages/stdlib";
+export type { Term } from "@packages/stdlib";
```
laggycomputer: Not fond of re-exports, but it's fine for now.

laggycomputer (changes requested):
Can we rename the migration to `enrollment_created_at_timestamp`?

laggycomputer:
Good to go, but I will not approve so that merging is blocked while we do our deployment to a separate worker.

apps/api/src/services/enrollment-history.ts
```diff
@@ -140,10 +146,17 @@ export class EnrollmentHistoryService {
...
+        sql`DATE(${websocSectionEnrollment.createdAt})`,
+        websocSectionEnrollment.sectionId,
+      ])
       .from(websocSectionEnrollment)
       .where(inArray(websocSectionEnrollment.sectionId, transformedSectionRows.keys().toArray()))
-      .orderBy(websocSectionEnrollment.createdAt);
+      .orderBy(
+        sql`DATE(${websocSectionEnrollment.createdAt})`,
+        websocSectionEnrollment.sectionId,
+        desc(websocSectionEnrollment.createdAt),
```
HwijungK: this clause is overshadowed by `DATE(created_at) ASC` clause no?

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +182,83 @@ export class EnrollmentHistoryService {
       }))
       .toArray();
   }
+
+  async getEnrollmentHistoryGranular(input: EnrollmentHistoryGranularServiceInput) {
+    const sectionRows = await this.db
```
HwijungK: could we seperate the querying for `sectionRows` into a helper function?

apps/api/src/services/enrollment-history.ts
HwijungK: Doesn't this get overshadowed by the ASC sort of DATE(created_at)? What purpose does this serve?

apps/api/src/rest/routes/enrollment-history.ts
HwijungK: Outdated doc?

apps/api/src/schema/enrollment-history.ts
```diff
@@ -65,3 +65,79 @@ export const enrollmentHistorySchema = z.object({
...
+    sectionCode: z
+      .string()
+      .regex(/^\d{5}$/, { error: "Invalid sectionCode provided" })
+      .transform((x) => Number.parseInt(x, 10))
+      .optional()
+      .openapi({ description: "The 5-digit section code", example: "36120" }),
+    sectionType: z
+      .enum(websocSectionTypes, { error: (_issue) => "Invalid sectionType provided" })
+      .optional(),
+    from: z.string().datetime({ error: "Invalid from date provided" }).openapi({
```
HwijungK: is having the start and end times as required strictly nessessary? Somebody might want to specify year/term/course and see all times, while we still give clients the option to query a specific time

apps/api/src/rest/routes/enrollment-history.ts
HwijungK: Or misplaced as its in the regular endpoint

apps/api/src/schema/enrollment-history.ts
HwijungK: use `z.iso.datetime()` instead

apps/api/src/rest/routes/enrollment-history.ts
```diff
@@ -31,6 +33,22 @@ const enrollmentHistoryRoute = createRoute({
...
 
+const enrollmentHistoryGranularRoute = createRoute({
+  summary: "Filter granular enrollment history",
+  operationId: "enrollmentHistoryGranular",
+  tags: ["Enrollment History"],
+  method: "get",
+  path: "/granular",
+  request: { query: enrollmentHistoryGranularQuerySchema },
+  description:
+    "Retrieves high-resolution enrollment history with full timestamps for the given parameters. Use from/to to scope the response to a specific time window.",
```
HwijungK: Should we add something about this data only being available after a certain date, as was the case for the general endpoint?

apps/api/src/services/enrollment-history.ts
aadi-shanker: I dont think so, my understanding with what I was doing is the first ORDER BY column matches the DISTINCT ON expression (DATE(createdAt)). The remaining columns are what determine which row Postgres picks within each distinct group. Then with DATE(createdAt) ASC, sectionId defines the grouping, and desc(createdAt) picks the latest snapshot of the day per section. Without this it would return an arbitrary row from each group. Its more clear once we integrate the multi snapshot data.

aadi-shanker: Duplicate question answered above

apps/api/src/rest/routes/enrollment-history.ts
aadi-shanker: I added 2 sentences regarding this let me know if that is clear enough.

HwijungK (approved):

laggycomputer (changes requested):
re-migrate

aadi-shanker:
> re-migrate

Done

apps/api/src/schema/enrollment-history.ts
```diff
@@ -65,3 +65,79 @@ export const enrollmentHistorySchema = z.object({
...
+      .string()
+      .regex(/^\d{5}$/, { error: "Invalid sectionCode provided" })
+      .transform((x) => Number.parseInt(x, 10))
+      .optional()
+      .openapi({ description: "The 5-digit section code", example: "36120" }),
+    sectionType: z
+      .enum(websocSectionTypes, { error: (_issue) => "Invalid sectionType provided" })
+      .optional(),
+    from: z.iso.datetime({ error: "Invalid from date provided" }).optional().openapi({
+      description: "Start of the time range (ISO 8601 timestamp). Optional.",
```
laggycomputer: No need to mark this as optional in text

apps/api/src/schema/enrollment-history.ts
```diff
@@ -65,3 +65,79 @@ export const enrollmentHistorySchema = z.object({
...
+  year: z.string(),
+  quarter: z.enum(terms),
+  sectionCode: z.string(),
+  department: z.string(),
+  courseNumber: z.string(),
+  sectionType: z.enum(websocSectionTypes),
+  sectionNum: z.string(),
+  units: z.string(),
+  instructors: z.string().array(),
+  meetings: z.object({ bldg: z.string().array(), days: z.string(), time: z.string() }).array(),
```
laggycomputer: I don't think there's a reason to serve these fields and anything besides identifying the section uniquely. Thoughts?

apps/api/src/services/enrollment-history.ts
```diff
@@ -138,12 +146,23 @@ export class EnrollmentHistoryService {
...
-      .select()
+      .selectDistinctOn([
+        sql`DATE(${websocSectionEnrollment.createdAt})`,
+        websocSectionEnrollment.sectionId,
+      ])
       .from(websocSectionEnrollment)
       .where(inArray(websocSectionEnrollment.sectionId, transformedSectionRows.keys().toArray()))
-      .orderBy(websocSectionEnrollment.createdAt);
+      .orderBy(
+        sql`DATE(${websocSectionEnrollment.createdAt})`,
```
laggycomputer: There's no need to convert this to DATE before ordering.

apps/api/src/services/enrollment-history.ts
```diff
@@ -138,12 +146,23 @@ export class EnrollmentHistoryService {
...
+        websocSectionEnrollment.sectionId,
+        desc(websocSectionEnrollment.createdAt),
```
laggycomputer: I'm not comfortable with two ORDER BYs like this without showing it's not a major performance issue. And shouldn't the snapshots be in increasing order anyway?

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +188,59 @@ export class EnrollmentHistoryService {
...
+        sectionNum: section.sectionNum,
+        units: section.units,
+        instructors: Array.from(section.instructors),
+        meetings: section.meetings.map(({ bldg, ...rest }) => ({
+          bldg: Array.from(bldg),
+          ...rest,
+        })),
+        finalExam: section.finalExam,
+        snapshots: [],
+      });
```
laggycomputer: See above about excess fields. That would probably also make this unnecessary.

Why can't we `ARRAY_AGG` this?

laggycomputer (changes requested):

apps/api/src/services/enrollment-history.ts
aadi-shanker: This requires DATE(createdAt) to lead the ORDER BY (Postgres constraint). The desc(createdAt) is the tiebreaker to pick the latest snapshot per day.

aadi-shanker:
I still need to take a look at the index query. For now here is the updated code for the existing comments

apps/api/src/services/enrollment-history.ts
```diff
@@ -138,12 +146,33 @@ export class EnrollmentHistoryService {
...
+  }
+
+  async getEnrollmentHistory(input: EnrollmentHistoryServiceInput) {
+    const sectionRows = await this.fetchSectionRows(input);
     const transformedSectionRows = transformSectionRows(sectionRows);
     const enrollmentRows = await this.db
-      .select()
+      .selectDistinctOn(
+        [sql`DATE(${websocSectionEnrollment.createdAt})`, websocSectionEnrollment.sectionId],
+        getTableColumns(websocSectionEnrollment),
```
laggycomputer: Wait, what? I don't see a reason to DISTINCT on every single enrollment datum. This is also probably going to be very inperformant.

apps/api/src/services/enrollment-history.ts
```diff
@@ -138,12 +146,33 @@ export class EnrollmentHistoryService {
...
+      )
+      .leftJoin(
+        websocInstructor,
+        eq(websocInstructor.name, websocSectionToInstructor.instructorName),
+      )
+      .where(buildQuery(input))
+      .orderBy(
+        sql`DATE(${websocSectionEnrollment.createdAt})`,
+        websocSectionEnrollment.sectionId,
+        desc(websocSectionEnrollment.createdAt),
```
laggycomputer: Earlier comments still stand. The most correct ordering is probably (section ID, ascending date). The current code also uses a similar ordering (note INCREASING created_at)

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +198,48 @@ export class EnrollmentHistoryService {
...
+      granularMapping.set(sectionId, {
+        year: section.year,
+        quarter: section.quarter,
+        sectionCode: section.sectionCode,
+        snapshots: [],
+      });
+    }
+    for (const row of enrollmentRows) {
+      const section = granularMapping.get(row.sectionId);
+      if (section) {
```
laggycomputer: Isn't this a logic error? Just throw if it isn't in the map...

laggycomputer (changes requested):
Are we still looking at precomputing last snapshot time and storing in `websoc_meta`?

aadi-shanker:
> Wait, what? I don't see a reason to DISTINCT on every single enrollment datum. This is also probably going to be very inperformant.

I think i need some clarification on this. My interpretation for the usage of DISTINCT on is so we can dedup the rows and only get one row?

My thought process
DISTINCT ON (DATE(createdAt), sectionId) tells Postgres: for each unique (date, sectionId) combination, give me only one row.

then grouped with
Combined with ORDER BY DATE(createdAt), sectionId, createdAt DESC, it picks the latest snapshot of each day for each section

That way we have our down sampling. Without DISTINCT on the query returns all snapshots per day?

LMK if my understanding is wrong but I am just confused.

laggycomputer:
> > Wait, what? I don't see a reason to DISTINCT on every single enrollment datum. This is also probably going to be very inperformant.
> 
> I think i need some clarification on this. My interpretation for the usage of DISTINCT on is so we can dedup the rows and only get one row?
> 
> My thought process DISTINCT ON (DATE(createdAt), sectionId) tells Postgres: for each unique (date, sectionId) combination, give me only one row.
> 
> then grouped with Combined with ORDER BY DATE(createdAt), sectionId, createdAt DESC, it picks the latest snapshot of each day for each section
> 
> That way we have our down sampling. Without DISTINCT on the query returns all snapshots per day?
> 
> LMK if my understanding is wrong but I am just confused.

I misread your code. Your DISTINCT ON is correct, I thought the enrollment columns were part of it. Sorry.

aadi-shanker:
> Earlier comments still stand. The most correct ordering is probably (section ID, ascending date). The current code also uses a similar ordering (note INCREASING created_at)

I believe the ORDER BY here is actually forced by Postgres. Because we do DISTINCT ON (X, Y), the ORDER BY has to start with those same expressions in the same order. We use DISTINCT ON (DATE(createdAt), sectionId), we can't lead with sectionId.

The desc(createdAt) is just a tiebreaker to pick the latest snapshot of each day just to clarify. So the data comes back in ascending date order because DATE(createdAt) ASC is the leading key then the JS loop dispatches each row to the right section anyway.

With the way it currently is I guess I can wrap this in a subquery to re-sort (sectionId, createdAt ASC) on the outer query  but, is it worth it?

apps/api/src/services/enrollment-history.ts
```diff
@@ -138,24 +147,43 @@ export class EnrollmentHistoryService {
...
+      )
+      .leftJoin(
+        websocInstructor,
+        eq(websocInstructor.name, websocSectionToInstructor.instructorName),
+      )
+      .where(buildQuery(input))
+      .orderBy(
+        websocSectionEnrollment.sectionId,
+        sql`DATE(${websocSectionEnrollment.createdAt})`,
+        desc(websocSectionEnrollment.createdAt),
```
laggycomputer: We've cleared up everything else, but I still think this should be ascending order.

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +197,46 @@ export class EnrollmentHistoryService {
...
+    const enrollmentConditions = [
+      inArray(websocSectionEnrollment.sectionId, transformedSectionRows.keys().toArray()),
+    ];
+    if (input.from) {
+      enrollmentConditions.push(gte(websocSectionEnrollment.createdAt, new Date(input.from)));
+    }
+    if (input.to) {
+      enrollmentConditions.push(lte(websocSectionEnrollment.createdAt, new Date(input.to)));
+    }
+    const enrollmentRows = await this.db
```
laggycomputer: Wait, isn't this two round trips? What's the motivation?

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -674,6 +724,9 @@ const doChunkUpsert = async (db: ReturnType<typeof database>, term: Term, resp:
       // the next scraping attempt will try another term first (if such another candidate term exists)
       // before retrying this one
       lastScraped: updatedAt,
+      ...(snapshotDecision.shouldSnapshot && {
+        lastSnapshot: snapshotDecision.scraperStartTime,
```
laggycomputer: `lastEnrollmentSnapshot`, perhaps?

packages/db/migrations/meta/_journal.json
```diff
@@ -232,6 +232,20 @@
...
+      "when": 1779237547865,
+      "tag": "0033_enrollment_created_at_timestamp",
+      "breakpoints": true
+    },
+    {
+      "idx": 34,
+      "version": "7",
+      "when": 1779992364084,
+      "tag": "0034_add_lastSnapshot_column",
+      "breakpoints": true
```
laggycomputer, line 241: Unify these migrations.

laggycomputer (changes requested):

apps/api/src/services/enrollment-history.ts
aadi-shanker: Yes it was intentional, the filter params (year, quarter, department, courseNumber) live on websocCourse, not on websocSectionEnrollment. My thought process was a single query would mean joining section -> course onto every enrollment row which is fine for the regular endpoint since DISTINCT ON collapses to one row per section/day, but the granular endpoint returns every snapshot (hundreds per section), so the join would balloon the row count before filtering. Splitting it: query 1 resolves the filter to section IDs, query 2 does WHERE sectionId IN (...) off the (sectionId, createdAt) index. There can be a consolidated version but this was done on purpose.

laggycomputer: > the join would balloon the row count before filtering

How does a join with exactly 1 row on the right side increase the row count?

> Splitting it: query 1 resolves the filter to section IDs, query 2 does WHERE sectionId IN (...) off the (sectionId, createdAt) index.

It is basically impossible for you to do this more performantly than the Postgres engine. The query planner understands that it can apply `WHERE` predicates to course or section rows before materializing the join with enrollment rows.

apps/api/src/schema/enrollment-history.ts
```diff
@@ -65,3 +65,71 @@ export const enrollmentHistorySchema = z.object({
...
   statusHistory: z.union([z.literal(""), z.enum(websocStatuses)]).array(),
 });
+
+export const enrollmentHistoryGranularQuerySchema = z
+  .object({
+    year: yearSchema.optional(),
+    quarter: z
+      .enum(terms, {
+        error: (issue) =>
+          issue.value === undefined ? "quarter is required" : "Invalid quarter provided",
```
laggycomputer: This error logic doesn't make sense if this is optional

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +197,46 @@ export class EnrollmentHistoryService {
...
+      .from(websocSectionEnrollment)
+      .where(and(...enrollmentConditions))
+      .orderBy(websocSectionEnrollment.createdAt);
+    const granularMapping = new Map<string, z.infer<typeof enrollmentHistoryGranularSchema>>();
+    for (const [sectionId, section] of transformedSectionRows) {
+      granularMapping.set(sectionId, {
+        year: section.year,
+        quarter: section.quarter,
+        sectionCode: section.sectionCode,
+        snapshots: [],
```
laggycomputer: Did you establish that this was slower in SQL? Even if it was, you could build the map from an iterator which would be more idiomatic and possibly faster.

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -860,6 +913,77 @@ async function scrapeGEsForTerm(db: ReturnType<typeof database>, term: Term) {
...
+    .from(calendarTerm)
+    .where(
+      and(
+        lte(calendarTerm.instructionStart, scraperTimestamp),
+        gte(calendarTerm.instructionEnd, scraperTimestamp),
+      ),
+    )
+    .limit(1);
+
+  // Get last snapshot scoped per-term to calculate time elapsed
```
laggycomputer: This comment isn't necessary, the correctness should speak for itself

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -925,9 +1050,9 @@ async function ingestChunk(
       `Chunk ${codeRangePretty} failed (probably too large); bisecting and trying again...`,
     );
 
-    const middle = lower + Math.floor((upper - lower) / 2);
-    await ingestChunk(db, term, lower, middle);
-    await ingestChunk(db, term, middle + 1, upper);
+    const middleInt = lower + Math.floor((upper - lower) / 2);
```
laggycomputer: Why this rename?

laggycomputer (changes requested):

apps/data-pipeline/websoc-scraper/src/lib.ts
aadi-shanker: My IDE is telling me I didnt change this but I will go ahead and put it back to middle.

apps/api/src/services/enrollment-history.ts
aadi-shanker: >How does a join with exactly 1 row on the right side increase the row count?

I thought about it and you are right I messed up the join semantics. It's many to one on the FK, so it wouldn't multiply enrollment rows. Will work on consolidating this query.

aadi-shanker: Consolidated to a single query. I took the enrollment query to filter itself through the course/section join via buildQuery, so the planner pushes the predicate down before touching enrollment like you said.

apps/api/src/services/enrollment-history.ts
aadi-shanker: This is now replaced based on the consolidation above. The map is now built from the result rows directly via Map.groupBy, which is the iterator-based approach you're describing.

apps/api/src/schema/enrollment-history.ts
aadi-shanker: Fixed, you commented on enrollmentHistoryGranularQuerySchema I did the same for the enrollmentHistoryQuerySchema.

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +196,51 @@ export class EnrollmentHistoryService {
...
+  async getEnrollmentHistoryGranular(input: EnrollmentHistoryGranularServiceInput) {
+    const enrollmentConditions = [buildQuery(input)];
+    if (input.from) {
+      enrollmentConditions.push(gte(websocSectionEnrollment.createdAt, new Date(input.from)));
+    }
+    if (input.to) {
+      enrollmentConditions.push(lte(websocSectionEnrollment.createdAt, new Date(input.to)));
+    }
+    const enrollmentRows = await this.db
+      .selectDistinctOn([websocSectionEnrollment.sectionId, websocSectionEnrollment.createdAt], {
```
laggycomputer: This DISTINCT ON is meaningless

apps/api/src/services/enrollment-history.ts
```diff
@@ -169,4 +196,51 @@ export class EnrollmentHistoryService {
       }))
       .toArray();
   }
+
+  async getEnrollmentHistoryGranular(input: EnrollmentHistoryGranularServiceInput) {
+    const enrollmentConditions = [buildQuery(input)];
```
laggycomputer: Why construct an array here?

apps/api/src/graphql/resolvers/enrollment-history.ts
```diff
@@ -8,5 +8,15 @@ export const enrollmentHistoryResolvers = {
       const service = new EnrollmentHistoryService(db);
       return await service.getEnrollmentHistory(enrollmentHistoryQuerySchema.parse(args.query));
     },
+    enrollmentHistoryGranular: async (
+      _: unknown,
+      args: { query: unknown },
+      { db }: GraphQLContext,
+    ) => {
+      const service = new EnrollmentHistoryService(db);
+      return await service.getEnrollmentHistoryGranular(
```
laggycomputer: Parse on the way out

apps/api/src/services/enrollment-history.ts
```diff
@@ -138,24 +146,43 @@ export class EnrollmentHistoryService {
       )
       .innerJoin(websocLocation, eq(websocLocation.id, websocSectionMeetingToLocation.locationId))
       .where(buildQuery(input));
+  }
+
+  async getEnrollmentHistory(input: EnrollmentHistoryServiceInput) {
+    const sectionRows = await this.fetchSectionRows(input);
```
laggycomputer: You're still fetching all section rows and incurring two round trips!

laggycomputer (changes requested):

apps/api/src/services/enrollment-history.ts
aadi-shanker: Removed fetchSectionRows entirely. Section metadata is now embedded as columns in the CTE alongside the enrollment data. The new single trip way is done by the CTE downsamples enrollment to one snapshot per section per day while selecting section metadata at the same time, then the outer query joins meetings/location onto the already-collapsed set. Code is more complex but I think its the right call.

laggycomputer (approved):

HwijungK (approved):
do we wanna look at how alpha table looks after fall add/drop or nah

laggycomputer:
> do we wanna look at how alpha table looks after fall add/drop or nah

yes we can look + merge conflicts

Issue #307, opened 2026-02-12
geText includes non GE text


andrew-wang0:
## Expected Behavior

The `ge_text` column in the `course` table should only include text if it relates to the GEs that the course fulfills.

Examples: `(III)` `(I and III)`

## Current Behavior

The check for valid `geText` [here](https://github.com/icssc/anteater-api/blob/06dec1c5b2dac9c79d347438247bf4b809a49da1/apps/data-pipeline/course-scraper/src/index.ts#L346) only validates if a string starts with `"("`. This can result in strings that do not describe the GEs of the course being counted as valid GE text.

Example: https://anteaterapi.com/v2/rest/courses/batch?ids=CBEMS225
Note geText is `(Design units: 0)`

## Possible Solution

Use more stringent GE text validation. 

Note that there are also oddities in the possible valid GE text strings, for example all of these are valid GE strings: `(II and Va.  ).` ,`(I)`, `(I, III)`, `(I and III)`, etc. These differences come from using the raw scraped string without cleaning.

PR #309, opened 2026-02-13
fix: check maybeGEText more

ZackeryHe:
## Description

geText sometimes has non GE text like `(Design units: 0)` or `(n/a. )`
now, geText will only be set if one of the regex for ge's matches

## Related Issue
https://github.com/icssc/anteater-api/issues/307 

## Notes

After a recrape, the 2 (and only) examples `(Design units: 0)` or `(n/a. )` still remain because the course is no longer listed in the department page. E.g CBEMS225 is no longer listed in its [course directory](https://catalogue.uci.edu/allcourses/cbems/). so won't be updated

Other example, UNISTU17C being `(n/a. )`. [old ahh course](https://antalmanac.com/planner/course/UNISTU17A)
## Types of changes

### DIFF
--- a/apps/data-pipeline/course-scraper/src/index.ts
+++ b/apps/data-pipeline/course-scraper/src/index.ts
@@ -344,36 +344,49 @@ function generateGEs(rawCourse: string[]) {
     geText: "",
   };
   if (!maybeGEText?.startsWith("(")) return res;
-  res.geText = maybeGEText;
-  if (res.geText.match(/\bI[Aa]\b/)) {
+  if (maybeGEText.match(/\bI[Aa]\b/)) {
     res.isGE1A = true;
   }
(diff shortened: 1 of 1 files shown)


apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -344,37 +344,50 @@ function generateGEs(rawCourse: string[]) {
     geText: "",
   };
   if (!maybeGEText?.startsWith("(")) return res;
```
sanskarm7:
```suggestion
  if (!possibleGEText?.startsWith("(")) return res;
```

sanskarm7 (approved):
tested well.. suppose there's no concrete way to find out with existing data, but new entries don't return problematic values, and seemingly maintain correct entries. well done!

PR #310, opened 2026-02-17
fix(degreeworks): Provide Polymorphic requirements for majors with specializations

### LINKED ISSUE #263: Account for polymorphic major requirement blocks
It's just been discovered that, similar to the way school requirement blocks change based on the selected major, it is also possible for major requirement blocks to change in response to ta choice of specialization.

The example observed is that the upper-division requirements for the B.S. in Applied and Computational Mathematics as of academic year 2025-2026 include a different, mutually exclusive set of upper-division requirements depending on which specialization/concentration (the words are used interchangeably here) is selected.
It is also possible to have no specialization for this major (#219).
There do exist three specializations named similarly to the major, but none of them produce a specialization block so they are not relevant, per changes in #218.
In essence, this happens:

* The upper-division requirements which take effect when no specialization applies are included in the major requirement block.
* Selection of the other specializtaions removes these requirements, **altering the major requirement block**, and adds the new requirements to the specialization block.

This means that storing one major block for all specializations in a major is no longer correct.

### LINKED ISSUE #308: degreeworks scraper: silly typo in ON CONFLICTs
In the DegreeWorks scraper, when we upsert majors, minors, and specializations, all three `ON CONFLICT`s are done with target `major.id`.
Some goober made a typo.

HwijungK:
## Description

We currently save one requirement for a major. However, when majors are taken with a specialization the original major block may change. For example, the Applied and Computational Math major has 2 optional specializations. The major requirement has a "Requirements for Pure Applied and Computation Math" subset, which is removed when specializations are taken.

We now run a major scrape for each major, spec pair. The `major_spec_pair_to_requirement` table refers to a major code and an optional potentially null spec code, and has a `requirement_id` that points to another table, `major_requirement`. To enforce uniqueness on major_requirement, a `requirements_hash` column is used as the json for requirements is to large to put a unique constrain on.

The `Major Requirement` Endpoint has been changed to take an optional `specializationId` (ex = '`BS-201A`) along with the `programId` (`BS-201`). It is an error to give specializationId as `null` or to give a specializationID that is not associated with the given majorId. 

## Related Issue

#263 

## How Has This Been Tested?

Locally tested on graphql and rest endpoints.

## Types of changes

## Checklist:


apps/api/src/schema/programs.ts
```diff
@@ -30,6 +30,10 @@ export const majorRequirementsQuerySchema = z.object({
     description: "A major ID to query requirements for",
     example: "BS-201",
   }),
+  specializationId: programIdBase.optional().openapi({
+    description: "fetch major requirement when taking specialization with this ID, if provided, ",
```
laggycomputer: Not specific enough, I think. 
> If provided, return the requirements for the major when taken with this specialization.

You should also add some note about the unspecified behavior if you don't put in a specialization for a major which requires one.

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -382,28 +399,43 @@ export class Scraper {
       }
     }
 
+    this.parsedPrograms = new Map([
+      ...this.parsedPrograms,
+      ...(await this.scrapePrograms(foundMajorSpecPairs)),
+    ]);
```
laggycomputer: mmm, this makes me uncomfy

can we just iterate over `await this.scrapePrograms` and push stuff in

laggycomputer (changes requested):

laggycomputer (changes requested):
You also accidentally checked two testing files in, no big deal

apps/api/src/schema/programs.ts
```diff
@@ -30,6 +30,11 @@ export const majorRequirementsQuerySchema = z.object({
     description: "A major ID to query requirements for",
     example: "BS-201",
   }),
+  specializationId: programIdBase.optional().openapi({
+    description:
+      "fetch major requirements when taking the specialization with this ID, if provided; providing no specialization when one is required returns the 'base' major requirements",
```
laggycomputer: > providing no specialization when one is required returns the 'base' major requirements"

Too nice of a claim; we can't really say anything about the meaning of the major requirements when a specialization is required, but none is provided...

laggycomputer (changes requested):

laggycomputer (changes requested):
closer

laggycomputer (changes requested):

packages/db/src/schema.ts
```diff
@@ -664,6 +669,34 @@ export const collegeRequirement = pgTable("college_requirement", {
   requirements: jsonb("requirements").$type<DegreeWorksRequirement[]>().unique().notNull(),
 });
 
+export const majorSpecPairToRequirement = pgTable("major_spec_pair_to_requirement", {
```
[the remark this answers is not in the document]
laggycomputer: Good point. Let's do what you suggest.

apps/data-pipeline/degreeworks-scraper/src/components/DegreeworksClient.ts
```diff
@@ -97,12 +97,14 @@ export class DegreeworksClient {
...
    * @param school this corresponds to the UCI notion of division, e.g. "U" or "G"
    * @param majorCode a major code
+   * @param specCode a specialization code
    * @param college this corresponds to the UCI notion of school, e.g. 55 for the school of bio sci
    */
   async getMajorAudit(
     degree: string,
     school: string,
     majorCode: string,
+    specCode?: string,
```
laggycomputer: More than one optional argument gets a little messy, especially if we add/reorder them. Do you want to pass in some kind of object instead? We could also leave this in if you don't care enough

laggycomputer (changes requested):

laggycomputer (changes requested):

apps/api/src/rest/routes/programs.ts
```diff
@@ -80,9 +80,7 @@ const majorRequirements = createRoute({
   tags: ["Programs"],
   method: "get",
   path: "/major",
-  description:
-    "Retrieve course requirements for a major in UCI's current catalogue. Note that these are the requirements for the major itself; " +
-    "if this major has specializations, then one is mandatory and its requirements apply as well.",
+  description: "Retrieve course requirements for a major in UCI's current catalogue.",
```
laggycomputer: Why did we remove the second part? It's still correct...

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -399,37 +418,42 @@ export class Scraper {
       }
     }
 
+    for (const [majorSpecId, majorProgram] of await this.scrapePrograms(foundMajorSpecPairs)) {
+      this.parsedPrograms.set(majorSpecId, majorProgram);
+    }
```
laggycomputer, line 426: We should have a comment here explaining briefly the premise of this issue and why we have to give these special attention.

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -159,25 +197,55 @@ async function main() {
...
+    const majorRequirementBlockIds = await tx
+      .insert(majorRequirement)
+      .values(majorRequirementBlocks)
+      .onConflictDoUpdate({
+        target: majorRequirement.id,
+        set: conflictUpdateSetAllCols(majorRequirement),
+      })
+      .returning({ id: majorRequirement.id })
+      .then((rows) => rows.map(({ id }) => id));
+
```
laggycomputer: This is very risky! The return order of these rows is not guaranteed to be the order you specified them.

laggycomputer (changes requested):
- rebase on main
- do we want to cache this stuff akin to the existing cache? not sure it's necessary myself but maybe it fits to you

apps/data-pipeline/degreeworks-scraper/src/index.ts
HwijungK: i don't think its nessessary to add a cache. Scraping the major spec pairs aren't a super significant part of the scrape because there's less specializations than majors already. plus those requirements are equally as important as the regular major requirements.

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -142,46 +154,39 @@ export class Scraper {
...
-      );
-
+  private async scrapePrograms(degrees: Iterable<ProgramCodes>) {
+    const ret = new Map<MajorSpecId, MajorProgram>();
+    for (const degree of degrees) {
+      const { collegeCode, majorCode, degreeCode, specCode } = degree;
+      const audit = await this.dw.getMajorAudit(degree);
       const majorAudit = audit?.major;
 
+      const specLogInfo = specCode ? ` specCode = ${specCode}` : "";
```
laggycomputer: for consistency
```suggestion
      const specLogInfo = specCode ? `, specCode = ${specCode}` : "";
```

apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
```diff
@@ -399,37 +418,44 @@ export class Scraper {
       }
     }
 
+    // Requirements in the majorBlock can change upon taking a specialization, so we must scrape the
+    // major block taken with each specialization.
```
laggycomputer: whitespace is negotiable, obviously
```suggestion
    // Requirements in the majorBlock can change upon taking a specialization, so we must rescrape programs now that we know the valid (major, specialization) combinations.
```

packages/db/src/schema.ts
```diff
@@ -669,6 +689,27 @@ export const collegeRequirement = pgTable("college_requirement", {
...
 });
 
+export const majorSpecializationToRequirement = pgTable(
+  "major_specialization_to_requirement",
+  {
+    majorId: varchar("major_id")
+      .notNull()
+      .references(() => major.id),
+    specializationId: varchar("specialization_id").references(() => specialization.id),
+    requirementId: bigint("requirement_id", { mode: "bigint" }).references(
```
laggycomputer: should this be not null?

laggycomputer (changes requested):

HwijungK:
Previous problem with major_spec_to_requirement not matching to an actual requirement happened because postgres wasn't picking up the unique index  on major and spec codes when spec was a null, leading to duplicate rows in major_specialization_to_requirement.

Changed how we match the generated hash from major_requirement and college_requirement to not be danergous and not use assertStrictEquals O(N^2) times.

Eventually, we should also change college_requirement to not have a randomly generated uuid.

apps/api/src/rest/routes/programs.ts
```diff
@@ -82,7 +82,7 @@ const majorRequirements = createRoute({
   path: "/major",
   description:
     "Retrieve course requirements for a major in UCI's current catalogue. Note that these are the requirements for the major itself; " +
-    "if this major has specializations, then one is mandatory and its requirements apply as well.",
+    "if this major has specializations and one is taken, its requirements apply as well",
```
laggycomputer:
```suggestion
    "if this major has specializations and one is taken, its requirements apply as well.",
```

laggycomputer (changes requested):

laggycomputer (changes requested):
As we discussed, get rid of `encodeBlock` and use the Postgres JSON hash to de-duplicate if necessary.

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
       requirements,
     }))
     .toArray();
+
   await db.transaction(async (tx) => {
+    await tx.execute(sql`SET CONSTRAINTS ALL DEFERRED`);
```
laggycomputer: Can we defer the minimum set of constraints possible to make this work? Principle of least action, etc...

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
       requirements,
     }))
     .toArray();
+
   await db.transaction(async (tx) => {
```
laggycomputer, line 86: Nit: call this `tx1`, so that we do not accidentally reference the outer transaction inside the inner transaction (which doesn't matter, technically, but it's a good practice)

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -43,40 +46,28 @@ async function main() {
...
+          majorId: `${degreeType}-${code}`,
+          degreeId: degreeType ?? "",
+          code,
+          ...(specCode !== undefined ? { specializationId: `${degreeType}-${specCode}` } : {}),
+          name,
+          specializationRequired,
+          requirements,
+          college,
+          ...(college ? { collegeReqUuid: uuidv4() } : {}),
+          majorReqUuid: uuidv4(),
```
laggycomputer: I'm going to call this sufficiently advanced technology that it is [indistinguishable from magic](https://en.wikipedia.org/wiki/Clarke's_three_laws), i.e. that we should comment the overall flow of what's going on in this transaction.

The steps I communicated to you privately are a good start, obviously better grammar would be desirable.

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
...
+          .from(majorRequirement)
+          .then((q) =>
+            q.map(({ uuid, hash }) => {
+              return [hash, uuid];
+            }),
+          ),
+      );
+      return new Map(
+        idToHash.map(({ uuid, hash }) => {
+          return [uuid, hashToUnified.get(hash)];
```
laggycomputer: Consider `getFromMapOrThrow` here (or just resolve if this really is meant to be optional).

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
...
+
+    const collegeReqUnifiedId = await tx.transaction(async (tx2) => {
+      const idToHash = await tx2
+        .insert(collegeRequirement)
+        .values(
+          majorSpecData
+            .filter((x) => x.college !== undefined)
+            .map(({ college, collegeReqUuid: uuid }) => {
+              return { name: college?.name, requirements: college?.requirements, id: uuid };
+            }) as (typeof collegeRequirement.$inferInsert)[],
```
laggycomputer: Why is this type hint necessary? There may be a good reason, but a cursory glance tentatively says otherwise

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
...
+          ),
+      );
+      return new Map(
+        idToHash.map(({ uuid, hash }) => {
+          return [uuid, hashToUnified.get(hash)];
+        }),
+      );
+    });
+
+    const collegeReqUnifiedId = await tx.transaction(async (tx2) => {
```
laggycomputer: Any way to somehow extract this into a (possibly type generic) helper function? This repeated logic makes me a big uncomfortable given its complexity.

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
...
+        }),
+      );
+    });
+
+    // Unifies randomly generated uuid associated with the same requirement hash
+    for (const majorSpecObj of majorSpecData) {
+      if (majorSpecObj.collegeReqUuid) {
+        majorSpecObj.collegeReqUuid = collegeReqUnifiedId.get(majorSpecObj.collegeReqUuid);
+      }
+      majorSpecObj.majorReqUuid = majorReqUnifiedId.get(majorSpecObj.majorReqUuid) as string;
```
laggycomputer: `getFromMapOrThrow` instead of casting!

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -93,7 +84,117 @@ async function main() {
...
+        idToHash.map(({ uuid, hash }) => {
+          return [uuid, hashToUnified.get(hash)];
+        }),
+      );
+    });
+
+    // Unifies randomly generated uuid associated with the same requirement hash
+    for (const majorSpecObj of majorSpecData) {
+      if (majorSpecObj.collegeReqUuid) {
+        majorSpecObj.collegeReqUuid = collegeReqUnifiedId.get(majorSpecObj.collegeReqUuid);
```
laggycomputer: This is a great place for a comment! Changing a UUID like this makes little sense out of context and in fact seems pathologically wrong if you don't understand (like you and I do) exactly what is happening and why. Give readers the benefit of that understanding.

laggycomputer (changes requested):

apps/data-pipeline/degreeworks-scraper/src/index.ts
HwijungK: .filtering out undefined still keeps `college` as a possibly undefined value during map, so i have to clarify that is not null

packages/db/src/schema.ts
```diff
@@ -661,12 +682,36 @@ export const schoolRequirement = pgTable("school_requirement", {
...
 export const collegeRequirement = pgTable("college_requirement", {
-  id: uuid("id").primaryKey().defaultRandom(),
+  requirementHash: bigint("requirement_hash", { mode: "bigint" })
+    .generatedAlwaysAs(sql`jsonb_hash_extended(requirements, 0)`) // also deferrable initially immediate
+    .unique()
+    .notNull(),
   name: varchar("name").notNull(),
   requirements: jsonb("requirements").$type<DegreeWorksRequirement[]>().notNull(),
-  requirementsHash: bigint("requirements_hash", { mode: "bigint" })
+  id: uuid("id").primaryKey().default(sql`gen_random_uuid()`),
```
laggycomputer: nit: drizzle gives you `defaultRandom` which we use elsewhere so it should be used here (and the second line where you do this)

probably will end up with no diff for `id` column here, since you can use the original definition again

laggycomputer (changes requested):
absolute cinema

apps/data-pipeline/degreeworks-scraper/package.json
```diff
@@ -13,6 +13,7 @@
     "@packages/stdlib": "workspace:*",
     "cross-fetch": "4.1.0",
     "jwt-decode": "4.0.0",
+    "uuid": "^13.0.0",
```
laggycomputer: we don't generate UUID anymore, right? can we remove this

laggycomputer (changes requested):
i gotta go do other stuff (play osu) rn but have this rq

apps/data-pipeline/degreeworks-scraper/package.json
HwijungK: uuid has been removed

laggycomputer (approved):
re-scrape before merging

we're done holy moly

PR #313, opened 2026-02-20
hotfix: doChunkUpsert for course updates all except GEs

sanskarm7:
dante you may not like this..

added util function `conflictUpdateSetAllColsExcept` that mirrors the existing `conflictUpdateSetAllCols` function.

this new function also takes a set of columns that **must** exist and removes them from the update list provided by the first parameter.

simply replaced problematic function call in `doChunkUpsert` with new one, and fed it a list of GEs (there may be a list somewhere to import, i couldn't remember)

### DIFF
--- a/apps/data-pipeline/websoc-scraper/src/lib.ts
+++ b/apps/data-pipeline/websoc-scraper/src/lib.ts
@@ -76,7 +76,7 @@ const geCategoryToFlag: Record<(typeof geCategories)[number], keyof CourseGEUpda
   "GE-8": "isGE8",
 };
 
-const geColumns = Object.values(geCategoryToFlag);
+const geColumns = Object.values(geCategoryToFlag) as string[];
 
 export async function getDepts(db: ReturnType<typeof database>) {
   const response = await fetch("https://www.reg.uci.edu/perl/WebSoc").then((x) => x.text());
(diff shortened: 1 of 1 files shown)


apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -50,6 +50,19 @@ const SECTIONS_PER_CHUNK = 891;
  */
 const LAST_SECTION_CODE = "97999";
 
+const GE_COLUMNS = [
```
laggycomputer: there's a map from category to column so go pull Object.values from that

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -50,6 +50,34 @@ const SECTIONS_PER_CHUNK = 891;
...
+  "GE-3": "isGE3",
+  "GE-4": "isGE4",
+  "GE-5A": "isGE5A",
+  "GE-5B": "isGE5B",
+  "GE-6": "isGE6",
+  "GE-7": "isGE7",
+  "GE-8": "isGE8",
+};
+
+const geColumns = Object.values(geCategoryToFlag);
```
laggycomputer: cast this to string[] instead of on use

laggycomputer (changes requested):

laggycomputer (approved):

Issue #275, opened 2025-12-02
Zod in DegreeWorks scraping


laggycomputer:
Let's add Zod to DegreeWorks response parsing and before storing data, so that we can check our assumptions going forward.
Silently sparse data or data in an unexpected shape has caused headaches in this area already.

lumakyns:
No worries!

PR #333, opened 2026-03-07
feat: add zod to degreeworks parsing

iAmbrosial:
## Description

Adds Zod runtime validation to all DegreeWorks API responses in the degreeworks scraper. Previously, API responses were cast directly to TypeScript types with no runtime checks. This meant malformed or unexpected responses would silently fail or lead to unclear downstream errors. We are thus introducing schemas for all response shapes and integrating them into their respective clients. 

Every variant of the `Rule` type has a `ruleType` field containing a unique string literal. To replicate this behaviour in an equivalent Zod schema, the `ruleTypeSchema` is defined as a **discriminated union**. Since discriminated unions cannot take intersections, every rule schema extends `ruleBaseSchema` to include the label field – reducing repetition if another rule type ever needs to be added. 

The `DWAuditErrorResponse` type denotes values that never occur (i.e., it is impossible for this field to have a valid value). It is thus nonsensical to validate the error response type with Zod, and we instead only create a `DWAuditOKSchema` and then directly validate the parsed results against that in the DegreeWorks Client. 

## Related Issue

#275 

## Motivation and Context

We are adding Zod to DegreeWorks response parsing and before storing data such that we can check our assumptions going forward. Without runtime validation, unexpected API response shapes and silently sparse data have caused issues in the area. Zod validation makes these failures visible immediately.

## How Has This Been Tested?

Ran the DegreeWorks Scraper for the 2025-2026 catalog year. The majority of undergraduate and graduate programs and all cached specializations parsed successfully. There were two cases of "Unexpected audit response shape" due to response objects lacking a blockArray field, but there seems to be no clear reason as to why other than DegreeWorks-side  errors. The one previously unclear ruleType was confirmed (hopefully) and added to ruleMarkerSchema.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/degreeworks-scraper/src/schema.ts
+++ b/apps/data-pipeline/degreeworks-scraper/src/schema.ts
@@ -14,6 +14,7 @@ export const withClauseSchema = z.object({
     "DWGRADETYPE",
     "DWPASSFAIL",
   ]),
+  connector: z.enum(["", "AND", "OR"]),
   operator: z.enum(["<", "<=", "=", ">", ">=", "<>"]),
   valueList: z.array(z.string()),
 });
@@ -150,6 +151,7 @@ export const blockSchema = z.object({
   requirementValue: z.string(),
   title: z.string(),
   ruleArray: z.array(ruleSchema),
+  catalogYear: z.string(),
 });
 
 // UndergraduateRequirements interface is not made into an equivalent Zod schema since it's made from already parsed Block objects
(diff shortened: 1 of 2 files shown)


laggycomputer (changes requested):
It seems we're now duplicating schemas and types, which is quite fragile. Can you remove every type which is now superseded by a Zod schema (or make them aliases to a `z.infer`)? I'm seeing type `WithClause` and `withClauseSchema`, for example.

laggycomputer (changes requested):
tiny stuff

laggycomputer (changes requested):
Totally slipped my mind that this was open while merging other things, so there are a handful of extra schemas to do now. Sorry about that...

PR #338, opened 2026-03-13
feat: serving "off"/closed dining periods

### LINKED ISSUE #332: Dining: serve "missing"/closed periods
As seen in test data [here](https://github.com/icssc/PeterPlate/blob/e2b70e5/packages%2Fapi%2Fsrc%2Fserver%2Fdaily%2FtestData%2FgetLocation.json#L4452), data is provided on when various schedules are "off", e.g. a regular schedule which does not take effect on a certain day.
Since we currently filter out such schedules/periods [here](https://github.com/icssc/anteater-api/blob/a04d1fb2d8cbd6f4c46130feaedebbfca55626b6/apps/data-pipeline/dining-scraper/src/update-restaurant.ts#L95), we don't serve data on "closed" periods, since they are considered irrelevant.
However, PeterPlate's Figma is calling for such closed periods to be served.
This is a novel extension not present in PeterPlate's original code.

This will require research to figoure out how exactly we can accomplish this.
Breaking changes are acceptable.

matt-franklin225:
## Description

UCI dining provides data on when various schedules are "off", which occurs when a certain meal period type is not served on a given day. Since the dining halls are closed on these days, we did not previously provide data for these periods. However, as mentioned in #332, PeterPlate's Figma is calling for such closed periods to be served.

To do this, the `parseOpeningHours` function in `util.ts` in `apps/data-pipeline/dining-scraper` has been adjusted such that, instead of skipping "off" periods, they are instead returned with a startTime and endTime of null (which is a breaking change, as these fields were previously not nullable and now are). Invalidly formatted date/time entries and empty ones will still be filtered out. In addition, under `apps/api/src/services/dining.ts`, the inner join onto diningDish using diningDish.id diningDishToPeriod.dishId was changed to a left join to allow for periods with no dishes to be returned.

## Related Issue

This will close #332 

## Motivation and Context

As previously mentioned, PeterPlate wants our API to return data on which days are closed, and these days are represented by the "off" periods.

## How Has This Been Tested?

The changes were initially tested locally to efficiently run through test cases, then was used for a standard run-through of the dining scraper to make sure it didn't break any previously working code. Also manually adjusted which schedules were being pulled (mixing between regular dining hours, summer times, and times when the dining hall is closed) to ensure that the adjustments would work both for the null cases and the standard ones.

## Types of changes

## Checklist:


matt-franklin225:
NOTE: we need to determine the migration number before merging

apps/data-pipeline/dining-scraper/src/util.ts
```diff
@@ -31,67 +31,41 @@ export function parseOpeningHours(hoursString: string): [WeekTimes, WeekTimes] {
...
-    if (timeRangeStr === "off") {
-      continue;
-    }
-
-    const [openTime, closeTime] = timeRangeStr.split("-"); // "07:15", "11:00"
-
-    if (!openTime || !closeTime) {
-      console.warn(`[parseOpeningHours]: Skipping block with incomplete time range: ${block}`);
-      continue;
+    // Only parse the range if it isn't "off"
```
laggycomputer: This comment has the wrong insight. It's pretty obvious to readers that we're filtering for strict inequality with `"off"`. But what does "off" *mean*?

laggycomputer (changes requested):
You can assume that your migration will be next, so you can keep 28 and promote out of draft. It's not like you could "predict" the next migration anyway, because the snapshot file has to be correct.

apps/api/src/schema/dining.ts
```diff
@@ -180,8 +180,8 @@ export const restaurantTodayResponseSchema = restaurantSchema.extend({
     z.string().openapi({ description: "The ID of a period." }),
     z.object({
       name: z.string().openapi({ description: "The name of a period", example: "Lunch" }),
-      startTime: z.iso.time(),
-      endTime: z.iso.time(),
+      startTime: z.iso.time().nullable(),
```
laggycomputer: Can you document what null means here?

laggycomputer (changes requested):

apps/data-pipeline/dining-scraper/src/util.ts
```diff
@@ -31,67 +31,41 @@ export function parseOpeningHours(hoursString: string): [WeekTimes, WeekTimes] {
...
+    if (timeRangeStr !== "off") {
```
laggycomputer, line 42: The string `""` also appears, and also indicates closure! Missing that here is a major correctness bug.

laggycomputer (changes requested):

apps/data-pipeline/dining-scraper/src/util.ts
matt-franklin225: Based on the way in which data is returned, we see "off" only on days when the dining hall is closed. We see the empty string if and only if the dining hall is open that day, but that particular meal is not being served (e.g. for Happy Hour during a regular day in either dining hall). It seemed unnecessary to return these segments, as we already have data for those days and our goal is to show which days are closed--unless our goal is to return the times or lack thereof for all meal types on all days. If we want to return those segments as well, I can adjust.

laggycomputer: > but that particular meal is not being served

This was the primary motivator for this PR. See PeterPlate Figma:
image

matt-franklin225: Updated and tested on my end, should work now

matt-franklin225: Actually I do see a couple more cases to account for, I'll work on those

apps/api/src/schema/dining.ts
matt-franklin225: Added a description, lmk if it's appropriately descriptive

apps/data-pipeline/dining-scraper/src/util.ts
matt-franklin225: As of now it works when I test on my end, adjusted the way in which it returns under src/services to allow it to return periods that have no dishes mapped to them which was the roadblock

matt-franklin225: What remains now is possibly returning null for cases where a period is served through some of a week but not all, I don't believe that currently works

matt-franklin225: All edge cases I know of are accounted for now

apps/api/src/schema/dining.ts
```diff
@@ -180,8 +180,14 @@ export const restaurantTodayResponseSchema = restaurantSchema.extend({
     z.string().openapi({ description: "The ID of a period." }),
     z.object({
       name: z.string().openapi({ description: "The name of a period", example: "Lunch" }),
-      startTime: z.iso.time(),
-      endTime: z.iso.time(),
+      startTime: z.iso.time().nullable().openapi({
+        description: "The start time of a period, or null if it doesn't happen that day",
```
laggycomputer: (repeat below)
```suggestion
        description: "The start time of a period, or null if not served on this day",
```

apps/data-pipeline/dining-scraper/src/update-restaurant.ts
```diff
@@ -90,10 +90,13 @@ export async function updateRestaurant(
...
-    const relevantMealPeriods = currentSchedule.mealPeriods.filter(
-      (mealPeriod) =>
-        mealPeriod.openHours[dayOfWeekToFetch] && mealPeriod.closeHours[dayOfWeekToFetch],
-    );
+    const relevantMealPeriods = currentSchedule.mealPeriods.filter((mealPeriod) => {
+      const open = mealPeriod.openHours[dayOfWeekToFetch];
+      const close = mealPeriod.closeHours[dayOfWeekToFetch];
+
+      // If either open or close time is empty, skip the meal period (null is allowed)
+      return !(open === "" || close === "");
```
laggycomputer: Is the time already split/parsed in `parseOpeningHours` to be either null or a meaningful time of day? Would this check ever trigger?

laggycomputer (changes requested):
Close to done...

apps/data-pipeline/dining-scraper/src/update-restaurant.ts
matt-franklin225: That's correct, this used to be relevant but no longer is with the new changes

apps/api/src/schema/dining.ts
```diff
@@ -180,8 +180,14 @@ export const restaurantTodayResponseSchema = restaurantSchema.extend({
...
     z.object({
       name: z.string().openapi({ description: "The name of a period", example: "Lunch" }),
-      startTime: z.iso.time(),
-      endTime: z.iso.time(),
+      startTime: z.iso.time().nullable().openapi({
+        description: "The start time of a period, or null if not served on this day",
+        example: "11:00:00",
+      }),
+      endTime: z.iso.time().nullable().openapi({
+        description: "The end time of a period, or null if it doesn't happen that day",
```
laggycomputer: Please make the revision you made above here as well.

laggycomputer (changes requested):

apps/data-pipeline/dining-scraper/src/update-restaurant.ts
laggycomputer: Can you revise this to say that the period dishes field is null? This comment becomes too ambiguous with the change you're making.

laggycomputer (changes requested):

laggycomputer (approved):
yuh

PR #344, opened 2026-03-29
feat(programs): multiple catalog years

### LINKED ISSUE #336: DegreeWorks: Scrape different catalogue years
The data we serve right now is an unspecified combination of various years scraped from various accounts over several months, over two academic years, over various entry years.
This mixing of catalogue years is quite dangerous because it corresponds to nobody's actual DegreeWorks requirements.
We need to serve different requirements per academic year.

laggycomputer:
- new table structure
- tables are prefixed with `dw_`
- optional catalogYear param to rest/graphql when requesting requirements; default behavior is unspecified but currently tries to use the latest catalog year of any program in the table, under the assumption that the request program will also be available for that year
- minor cleanups on graphql around optionality of queries
- ~~**still need endpoints to show list of available years**~~

## test

- all the graphql and rest endpoints with catalog options included and excluded


apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -86,135 +83,180 @@ async function main() {
...
           {
             id: "GE",
+            catalogYear,
             requirements: geRequirementData,
           },
         ])
         .onConflictDoUpdate({
-          target: schoolRequirement.id,
-          set: conflictUpdateSetAllCols(schoolRequirement),
+          target: dwSchoolRequirement.id,
```
HwijungK: target should target `[id, catalogYear]` in all inserts to `dwSchoolRequirement`

apps/data-pipeline/degreeworks-scraper/src/index.ts
```diff
@@ -86,135 +83,180 @@ async function main() {
...
-        target: specialization.id,
-        set: conflictUpdateSetAllCols(specialization),
+        target: [dwMajorYear.programId, dwMajorYear.catalogYear],
+        set: conflictUpdateSetAllCols(dwMajorYear),
       });
 
     await tx
-      .insert(majorSpecializationToRequirement)
-      .values(majorSpecToRequirementData)
+      .insert(dwMajorSpecializationToRequirement)
```
HwijungK, line 212: Cannot insert into this table before adding rows into `dwSpecialization` It will cause a foreign key constraint violation!

apps/api/src/schema/programs.ts
```diff
@@ -4,18 +4,26 @@ const programIdBase = z.string({
   error: (issue) => (issue.input === undefined ? "programId is required" : "invalid programId"),
 });
 
+const catalogYearInputSchema = z.string().optional().openapi({
+  description:
+    "If specified, data is derived from the closest known catalog year (prioritizing more recent years in case of a tie); otherwise, data is from an unspecified year",
```
HwijungK: `most recent year` 🤓

More importantly, Can i make the argument that we should provide the most recent found catalogue year when unspecified. That feels like the logical behaviour.

apps/api/src/schema/programs.ts
HwijungK, line 50: Add that catalogYears has no know effect on specialization for consistency if true.

HwijungK (changes requested):

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,102 @@ import type {
 export class ProgramsService {
   constructor(private readonly db: ReturnType<typeof database>) {}
 
+  private catalogYearModifiers(
```
HwijungK: We might want a comment to doc what this function does

apps/api/src/services/programs.ts
```diff
@@ -105,60 +153,90 @@ export class ProgramsService {
...
-          eq(major.id, majorSpecializationToRequirement.majorId),
-        )
-        .leftJoin(
-          majorRequirement,
-          eq(majorSpecializationToRequirement.requirementId, majorRequirement.id),
-        )
-        .limit(1);
-      return orNull(got);
+        );
+      const [got] = await (order !== undefined ? base.orderBy(order) : base).limit(1);
```
HwijungK, line 226: not strictly need to `limit` since `[got]` gets the first element anyway iirc.

HwijungK:
image
Seems like `catalogYear` return field is incorrect?  This is with both 20242025 and 20252026 catalogs in local database

apps/api/src/services/programs.ts
laggycomputer: limit can still help the optimizer in general (not here) and prevents unnecessary sending of rows to be discarded immediately

apps/api/src/schema/programs.ts
laggycomputer: > most recent year

that is not the behavior

> we should provide the most recent found catalogue year when unspecified

[no](https://canary.discord.com/channels/772739905981644850/910727722530123797/1506461902660702208)

apps/api/src/services/programs.ts
```diff
@@ -105,60 +153,90 @@ export class ProgramsService {
...
+        .innerJoin(dwMajorYear, eq(dwMajor.id, dwMajorYear.programId))
         .where(
           and(
-            eq(major.id, query.programId),
+            eq(dwMajor.id, query.programId),
             query.specializationId
-              ? eq(majorSpecializationToRequirement.specializationId, query.specializationId)
+              ? eq(dwMajorSpecializationToRequirement.specializationId, query.specializationId)
               : undefined,
+            where,
```
HwijungK: We filter by exact client given catalog year here, which means that if the client enters  a catalog year we haven't saved, `base` has length of 0 and we are unable to ever find the catalog of nearby years.

HwijungK (changes requested):

apps/api/src/schema/programs.ts
laggycomputer: Currently, it doesn't, but it depends. Every specialization has a major, since we just infer associated major from code, but as specializations are retired or created they will have sparse data over many years. Should we only include specs with requirements here? Thoughts?

HwijungK: I feel like we should have an optional catalogYears field which will only fetch valid specializations from that catalog year. Otherwise, we could show all specializations that have existed.

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,101 @@ import type {
...
+      .innerJoin(dwMajor, eq(majorSpecialization.id, dwMajor.id))
+      .innerJoin(dwDegree, eq(dwMajor.degreeId, dwDegree.id))
+      .innerJoin(
+        dwMajorYear,
+        and(
+          eq(dwMajor.id, dwMajorYear.programId),
+          eq(majorSpecialization.catalogYear, dwMajorYear.catalogYear),
+        ),
+      )
+      .where(and(query.id ? eq(majorSpecialization.id, query.id) : undefined));
```
HwijungK: unnecessary use of `and`

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,101 @@ import type {
...
+    const majorSpecialization = this.db
+      .$with("major_specialization")
+      .as(
+        (specsOrder !== undefined
+          ? majorSpecializationInner.orderBy(specsOrder)
+          : majorSpecializationInner
+        ).groupBy(dwMajor.id, dwMajorSpecializationToRequirement.catalogYear),
+      );
 
     return this.db
```
HwijungK: Currently gives entries for all catalog years we have, instead of just  the matched one.

Also the orderBy clause seems to not be having an effect?
```sql
With query as (
  SELECT '20252026' AS "year"
)
SELECT dw_major.id as id, dw_major.name as "name", msr.catalog_year as catalogYear, ARRAY_REMOVE(ARRAY_AGG(msr.specialization_id), NULL) as "specializations"
FROM dw_major
LEFT JOIN dw_major_specialization_to_requirement as msr
ON dw_major.id = msr.major_id
GROUP BY (dw_major.id, msr.catalog_year)
ORDER BY (
  SELECT
  CASE
  WHEN msr.catalog_year > query.year THEN SUBSTRING(msr.catalog_year FROM 1 FOR 4)::real - SUBSTRING(query.year FROM 1 FOR 4)::real
  WHEN msr.catalog_year = query.year THEN 0
  WHEN msr.catalog_year < query.year THEN SUBSTRING(query.year FROM 1 FOR 4)::real - SUBSTRING(msr.catalog_year FROM 1 FOR 4)::real + .5
  END
  FROM query
)
```

This seems to work. But on the staging endpoint, catalogYear field has no effect on the returned order of entries.

apps/api/src/services/programs.ts
```diff
@@ -105,60 +152,86 @@ export class ProgramsService {
...
           schoolRequirements: {
-            name: collegeRequirement.name,
-            requirements: collegeRequirement.requirements,
+            name: dwMajorYear.collegeRequirements,
+            requirements: dwMajorYear.collegeRequirementsTitle,
           },
         })
-        .from(major)
+        .from(dwMajor)
+        .innerJoin(
```
HwijungK, line 207: inner joining `dwMajorRequirement` and `dwMajorYear` like this means that rows from major requirement table and major year table with DIFFERENT catalog years will be joined together!

laggycomputer:
@HwijungK This probably warrants a new round of testing for the new semantics...

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,132 @@ import type {
...
+      .where(
+        and(
+          query.id ? eq(dwMajor.id, query.id) : undefined,
+          query.catalogYear
+            ? inArray(
+                dwMajorYear.catalogYear,
+                this.db
+                  .select({ best: dwMajorYear.catalogYear })
+                  .from(dwMajorYear)
+                  .where(and(query.id ? eq(dwMajorYear.programId, query.id) : undefined))
```
HwijungK: Still an unnecessary `and` .

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,132 @@ import type {
...
+                  .orderBy(
+                    this.catalogYearPriorityExpression(
+                      dwMajorYear,
+                      query.catalogYear,
+                    ) as SQL<unknown>,
+                  )
+                  .limit(1)
+                  .as("best_year"),
```
HwijungK: Apparently, it is illegal to name subqueries used in where:

ex:
```

SELECT * FROM dw_major_year
WHERE catalog_year in (
  (SELECT catalog_year FROM dw_major_year) AS sub
)
```
fails

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,132 @@ import type {
...
-    const majorSpecialization = this.db.$with("major_specialization").as(
-      this.db
-        .select({
-          id: major.id,
-          name: major.name,
-          specializations: sql`ARRAY_REMOVE(ARRAY_AGG(${specialization.id}), NULL)`.as(
+    return await this.db
+      .select({
+        id: dwMajor.id,
+        name: dwMajor.name,
```
HwijungK, line 61: Cannot select because it isn't part of GroupBy or in an Aggregation

HwijungK (changes requested):

apps/api/src/services/programs.ts
HwijungK: can't alias

HwijungK, line 133: is checking majorId b/c we're preparing for a case when a specialization could be taken by multiple majors. I don't remember if we ever discussed that.

Otherwise, checking for specializationId should be enought?

HwijungK (changes requested):

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,137 @@ import type {
...
+        and(
+          query.majorId ? eq(dwSpecialization.majorId, query.majorId) : undefined,
+          query.catalogYear
+            ? inArray(
+                dwMajorSpecializationToRequirement.catalogYear,
+                this.db
+                  .select({ best: dwMajorSpecializationToRequirement.catalogYear })
+                  .from(dwMajorSpecializationToRequirement)
+                  .where(
+                    and(
```
HwijungK: unecessary use of and...

apps/api/src/services/programs.ts
```diff
@@ -28,56 +30,137 @@ import type {
...
+      .where(query.id ? eq(dwMinor.id, query.id) : undefined);
   }
 
   async getSpecializations(query: z.infer<typeof specializationsQuerySchema>) {
     return this.db
-      .select({
-        id: specialization.id,
-        majorId: specialization.majorId,
-        name: specialization.name,
+      .selectDistinct({
```
HwijungK: why selectDistinct, would there ever be a case where duplicates are given?

apps/api/src/services/programs.ts
```diff
@@ -105,60 +188,92 @@ export class ProgramsService {
...
-          eq(major.id, majorSpecializationToRequirement.majorId),
-        )
-        .leftJoin(
-          majorRequirement,
-          eq(majorSpecializationToRequirement.requirementId, majorRequirement.id),
-        )
-        .limit(1);
-      return orNull(got);
+        );
+      const [got] = await (order !== undefined ? base.orderBy(order) : base).limit(1);
```
HwijungK, line 226: If we query for a specialization that was removed with the wrong year (say in the future we call `getMajorRequirements` with `catalogYear=20262027` and `specializationId=BS-201A` when specs no longer exist for the CS major) we will give them the closest catalogYear that included that spec (20252026) even if the year that was queried for exists in our records.

Is this a behaviour we want? it seems dangerous imo.

apps/api/src/services/programs.ts
```diff
@@ -105,60 +188,92 @@ export class ProgramsService {
...
-          requirements: majorRequirement.requirements,
+          id: dwMajor.id,
+          name: dwMajor.name,
+          catalogYear: dwMajorYear.catalogYear,
+          requirements: dwMajorRequirement.requirements,
           schoolRequirements: {
-            name: collegeRequirement.name,
-            requirements: collegeRequirement.requirements,
+            name: dwMajorYear.collegeRequirements,
+            requirements: dwMajorYear.collegeRequirementsTitle,
```
HwijungK: switched `name` and `requirement`

HwijungK (changes requested):

apps/api/src/services/programs.ts
laggycomputer: Resolved elsewhere; since the list of specializations for a major is also year-dependent after this PR, this is fine

apps/api/src/services/programs.ts
laggycomputer: you right this is probably an artifact of something

packages/db/src/schema.ts
```diff
@@ -722,82 +722,115 @@ export const instructorToWebsocInstructor = pgTable(
...
-export const schoolRequirement = pgTable("school_requirement", {
-  id: varchar("id").primaryKey(),
-  requirements: json("requirements").$type<DegreeWorksRequirement[]>().notNull(),
-});
+export const dwSchoolRequirement = pgTable(
+  "dw_school_requirement",
+  {
+    id: varchar("id").notNull(),
+    catalogYear: varchar("catalog_year").notNull(),
+    requirements: json("requirements").$type<DegreeWorksRequirement[]>().notNull(),
```
HwijungK: do we use jsonb?

apps/api/src/schema/programs.ts
```diff
@@ -5,25 +5,42 @@ const programIdBase = z.string({
   error: (issue) => (issue.input === undefined ? "programId is required" : "invalid programId"),
 });
 
+const requirementsCatalogYearInputSchema = z.string().optional().openapi({
```
HwijungK: nit can we add a `refine`. pmo when i can say `20182025` and it workss

apps/api/src/services/programs.ts
```diff
@@ -105,60 +186,92 @@ export class ProgramsService {
...
+          eq(dwMajorSpecializationToRequirement.requirementId, dwMajorRequirement.id),
+        )
         .where(
           and(
-            eq(major.id, query.programId),
+            eq(dwMajor.id, query.programId),
             query.specializationId
-              ? eq(majorSpecializationToRequirement.specializationId, query.specializationId)
+              ? eq(dwMajorSpecializationToRequirement.specializationId, query.specializationId)
               : undefined,
```
HwijungK: if a specializationId was not given for a major with optional specs, we must ensure that we return the requirements with no spec associated.

apps/api/src/services/programs.ts
```diff
@@ -121,7 +121,7 @@ export class ProgramsService {
 
   async getSpecializations(query: z.infer<typeof specializationsQuerySchema>) {
     return this.db
-      .selectDistinct({
+      .select({
```
HwijungK: wait i was wrong. selectDistinct is required b/c we join with majorspectoreq, which adds duplicates for each catalogYear

apps/api/src/schema/programs.ts
```diff
@@ -5,25 +5,42 @@ const programIdBase = z.string({
...
+  description:
+    "If specified, data is derived from the closest known catalog year (prioritizing more recent years in case of a tie); otherwise, data is from an unspecified year",
+  example: "20242025",
+});
+
+const majorCatalogYearInputSchema = requirementsCatalogYearInputSchema.openapi({
```
HwijungK: Doesn't seem necessary, we only use it 2 times and one of them we override the description anyway which defeats the purpose

HwijungK (changes requested):
Very close

packages/db/src/schema.ts
laggycomputer: we actually didn't here but let's just make it the case

HwijungK (approved):
lgtm. Congrats on another gem from the mines!

PR #357, opened 2026-04-08
feat(degreeworks): serve dwcredit requirements

### LINKED ISSUE #337: DegreeWorks: Serve DWCredit requirements
background: 

Currently, we use DWCredits to check if it is theoretically possible for a course to satisfy constraint unit requirements. For courses with variable units, whether the course will satisfy the requirements depends on the number of units the user chooses. We should instead return DWCredits requirement as part of our requirements endpoint.

see #286 for original implementation, and #322 for how DWrequirements are served.

ZackeryHe:
## Description

serve dwcredits
## Related Issue
https://github.com/icssc/anteater-api/issues/337

## Types of changes

## Checklist:


apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -327,7 +329,16 @@ export class AuditParser {
               ),
             ).then((x) =>
               x
-                .flatMap(([classes, withArray]) => this.filterThroughWithArray(classes, withArray))
+                .flatMap(([classes, withArray]) => {
+                  const filtered = this.filterThroughWithArray(classes, withArray);
```
HwijungK: This will filter out all the courses that definatley can't be taken for this requirement, but not those that definately can be taken, leading to constraints that are always fulfilled

apps/api/src/schema/programs.ts
```diff
@@ -52,6 +52,23 @@ export const ugradRequirementsQuerySchema = z.object({
   id: z.enum(UgradRequirementsBlockIds).openapi({ description: "The requirements block to fetch" }),
 });
 
+const unitConstraintSchema = z.object({
+  type: z.literal("unit"),
+  connector: z.enum(["", "AND", "OR"]),
+  operator: z.string(),
```
HwijungK: can this be an enum?

HwijungK (changes requested):

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
ZackeryHe: added strict filter `isUnitConstraintAlwaysSatisfied`
Unfortunately, don't think we can combine into a helper with either of the other functions that use switch

HwijungK: making `parseUnitRequirement` return a function was bad design on my end. Lets have that return a bool.

Also, to make `parseUnitRequirement` more general, could we replace the `maxUnit` `minUnit` parameters with `unit`, and check if `.some` `unit` in a range between the minUnit and maxUnit can be applied (for sometimes applied courses) or if `.all` possible units apply (in which case we don't add a unit constraint to it as it would be redundant)

HwijungK (changes requested):
see comment

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
ZackeryHe: also collapsed DWTERM stuff into the same `satisfies()` function following this logic

ZackeryHe:
I think we might have to serve the DWTERM constraint eventually as well. e.g DWTERM > FALL 2025. technically passes for catalog year 2025-2026. However, consumers would need to know you still can't take it in the fall

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -200,58 +184,75 @@ export class AuditParser {
...
     }
   }
 
+  private possibleUnitsForCourse(c: typeof course.$inferSelect): number[] {
+    const minUnits = Number.parseInt(c.minUnits, 10);
+    const maxUnits = Number.parseInt(c.maxUnits, 10);
+    return Array.from({ length: maxUnits - minUnits + 1 }, (_, i) => minUnits + i);
+  }
+
+  private isUnitConstraintAlwaysSatisfied(
```
HwijungK: I don't think this had to become its own function. and could instead be implemented in `classMatchesWithClause`

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -200,58 +184,75 @@ export class AuditParser {
...
-        );
-        // valueList contains specific term(s) from the DWTERM constraint
-        // Courses pass this filter only if at least one term in the current
-        // catalog year satisfies a DWTERM predicate
+        // valueList contains specific term(s) from the DWTERM constraint.
+        // Include the course if some term in the current catalog year satisfies
+        // some target term predicate (i.e. the course is offered in a matching
+        // term during the catalog year).
+        const terms = AuditParser.schoolYearTerms(this.catalogYear);
+        // For <>, require every term in the school year to satisfy t !== target
```
HwijungK: Is there a reason that <> has to be fullfilled by every term but not the other operators? or can it be taken out like dwcredit (I don't know that much about dwterm so)?

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -200,58 +184,75 @@ export class AuditParser {
...
     switch (withClause.code) {
       case "DWCREDIT":
-      case "DWCREDITS":
-        return this.parseUnitRestrictionOperator(withClause.operator)(
-          Number.parseInt(c.minUnits, 10),
-          Number.parseInt(c.maxUnits, 10),
-          withClause.valueList,
-        );
+      case "DWCREDITS": {
+        // for > and >= operations, we use a loose interpretation for variable unit courses
```
HwijungK: I believe the strict vs. loose interpretation is needed, but should not be determined through the operators.

Previously, we were choosing either to include or exclude the courses, without serving a more nuanced constraint. We chose to serve the course if it was POSSIBLE to take the class and have it apply for the requirement. if it was NEVER possible, that course would be filtered out. that's why >, >=, and = have a loose interpretation. These operators are only found in the included course list, so we want to serve them if it is possible to take those courses for credit (loose interpretation, that is).

 the < and <= operators were only found in the excluded course list. If a class ends up on the excluded course list, than that course is removed from the final course list. (final course list = includedCourses - excludedCourses). we want to remove the course from the final course list ONLY IF that course will ALWAYS be excluded (i.e. all possible units for that course matches the withClause -> all possible units for that class would result it in being in excluded courses -> that course would be excluded from the final course list). This is why those operators follow the strict interpretations.

The goal of this PR is to add a unitConstraint for courses that follow a loose interpretation (i.e it is possible with SOME units but not others). Additonally we could filter out courses that will NEVER meet the withArray constraints, and not include constraints for courses that ALWAYS meet the constrait (this part isn't strictly nessessary as all the correct data will be given anways, it just removes redundancy. If you choose not to include this part it's ok).

TLDR: There may be a need to call this method in a strict/loose mode, but that shouldn't be determined by the operator.

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
HwijungK, line 224: how should we handle a withClause in the excluded courses?

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
HwijungK: maybe we could add a `strict` boolean parameter on `classMatchesWithClause` instead?.

HwijungK (changes requested):

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
ZackeryHe: tbh its all guessing rn. the only DWTERM I've seen is from fall. After thinking again, <> should probably be loose as well.

apps/data-pipeline/degreeworks-scraper/src/components/WithArrayUtils.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  return { type, children };
+}
+
+function clauseToLeaf(clause: WithClause): CourseConstraintTree {
+  return {
+    type: "leaf",
+    code: clause.code,
+    operator: clause.operator,
+    valueList: clause.valueList,
```
HwijungK, line 93: use ...clause shorthand

apps/data-pipeline/degreeworks-scraper/src/components/WithArrayUtils.ts
```diff
@@ -0,0 +1,228 @@
...
+  }
+  if (currentGroup.length > 0) {
+    orGroups.push(currentGroup);
+  }
+  return orGroups;
+}
+
+function evaluateLeaf(
+  c: typeof course.$inferSelect,
+  leaf: Extract<CourseConstraintTree, { type: "leaf" }>,
```
HwijungK: Not too familiar with Extract<> but is this any different from simply having it type `CourseConstraintLeaf`?

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -358,6 +264,20 @@ export class AuditParser {
                 : this.lexOrd(a.department, b.department),
             )
             .map(([x]) => x);
+
+          const courseConstraints: Record<string, CourseConstraintTree> = {};
+          for (const id of courses) {
+            const tree = andTrees([
+              orTrees(includeTreesById.get(id) ?? []),
```
HwijungK, line 271: should this not be andTrees?

HwijungK (commented):
Remember to fix merge conflict

apps/data-pipeline/degreeworks-scraper/src/components/WithArrayUtils.ts
ZackeryHe: clause has connector which we don't want

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
ZackeryHe: yeah this is `andTrees`, but i first combine them with `orTrees` assuming the single course can show up more than once in the block.

HwijungK: That fair

packages/db/src/schema.ts
```diff
@@ -108,16 +108,47 @@ export type DegreeWorksProgram = DegreeWorksProgramId & {
...
+ */
+export type CourseConstraint = {
+  code:
+    | "DWCREDIT"
+    | "DWCREDITS"
+    | "DWTERM"
+    | "DWLOCATION"
+    | "DWTITLE"
+    | "DWGRADETYPE"
+    | "DWPASSFAIL";
```
HwijungK: lets extract this into an enum and so we can import it into schema/programs.ts

apps/data-pipeline/degreeworks-scraper/src/components/WithArrayUtils.ts
```diff
@@ -0,0 +1,228 @@
...
+    throw new Error("Expected exclusion inversion to contain at least one clause");
+  }
+  if (withArray.length > 1) {
+    throw new Error(
+      `Expected exclusion inversion to contain exactly one clause, got ${withArray.length}`,
+    );
+  }
+
+  const [clause] = withArray;
+  if (clause.code !== "DWCREDIT" && clause.code !== "DWCREDITS") {
```
HwijungK, line 53: Lets leave a comment explaining that we could implement more withClauses in the future here

HwijungK (approved):

HwijungK:
Its mostly good I think. Just nitpicking now

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -371,6 +289,7 @@ export class AuditParser {
               requirementType,
               courseCount: Number.parseInt(rule.requirement.classesBegin, 10),
               courses,
+              ...(hasCourseConstraints && { courseConstraints }),
```
laggycomputer: If `hasCourseConstraints` just checks that the object is not empty, why not just spread the constraints unconditionally?

packages/db/src/schema.ts
```diff
@@ -108,16 +108,53 @@ export type DegreeWorksProgram = DegreeWorksProgramId & {
...
+ */
+export enum WithConstraintCode {
+  DWCREDIT = "DWCREDIT",
+  DWCREDITS = "DWCREDITS",
+  DWTERM = "DWTERM",
+  DWLOCATION = "DWLOCATION",
+  DWTITLE = "DWTITLE",
+  DWGRADETYPE = "DWGRADETYPE",
+  DWPASSFAIL = "DWPASSFAIL",
+}
```
laggycomputer: We try to stick with the strippable subset of TS so make this an object `as const`.

apps/api/src/schema/programs.ts
```diff
@@ -52,6 +53,29 @@ export const ugradRequirementsQuerySchema = z.object({
...
+
+const courseConstraintTreeSchema: z.ZodType<CourseConstraintTree> = z.union([
+  courseConstraintLeafSchema,
+  z.object({
+    type: z.enum(["AND", "OR"]),
+    children: z.lazy(() => courseConstraintTreeSchema).array(),
+  }),
+]);
+
+const courseConstraintsField = z.record(z.string(), courseConstraintTreeSchema).optional().openapi({
```
laggycomputer: Schema names should end with `schema` (in camelCase as appropriate), so maybe this should be called `courseConstraintsSchema`.

laggycomputer (changes requested):
New framework looks impressive. Nits below.

apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
ZackeryHe: Changed it a little bit. I don't think `{ ...courseConstraints }` works bc it will emit `{}` when no courseConstraints

apps/api/src/schema/programs.ts
```diff
@@ -52,6 +53,32 @@ export const ugradRequirementsQuerySchema = z.object({
...
+    children: z.lazy(() => courseConstraintTreeSchema).array(),
+  }),
+]);
+
+const courseConstraintsSchema = z
+  .record(z.string(), courseConstraintTreeSchema)
+  .optional()
+  .openapi({
+    description:
+      "A map from course ID to the boolean expression tree of constraints for that course.",
```
laggycomputer: Can we specify the semantics of a missing value?

laggycomputer (changes requested):

packages/db/src/schema.ts
```diff
@@ -129,16 +129,56 @@ export type MajorProgram = {
...
+ * We use a runtime value so zod schema can consume it
+ */
+export const WithConstraintCode = {
+  DWCREDIT: "DWCREDIT",
+  DWCREDITS: "DWCREDITS",
+  DWTERM: "DWTERM",
+  DWLOCATION: "DWLOCATION",
+  DWTITLE: "DWTITLE",
+  DWGRADETYPE: "DWGRADETYPE",
+  DWPASSFAIL: "DWPASSFAIL",
```
laggycomputer: Hmm, this doesn't look that great. Let's just use an array of strings, then?

packages/db/src/schema.ts
```diff
@@ -129,16 +129,56 @@ export type MajorProgram = {
...
+} as const;
+
+export type WithConstraintCode = keyof typeof WithConstraintCode;
```
laggycomputer: Then this would have to be `typeof WithConstraintCode[number]`

laggycomputer (changes requested):

laggycomputer (approved):

PR #361, opened 2026-04-09
fix: study room dependent on session tz

### LINKED ISSUE #360: Study room timezone inconsistencies
We have reports of multiple calls of the same endpoint returning no data or different data.
Timezone shifting may be at fault, i.e. incorrect consideration of local time when processing input (likely not output).
It has been suggested that we move all study room times to use `TIMESTAMP WITH TIME ZONE`.

ZackeryHe:
## Description
currently depends on session tz

Also, locked response to UTC, was helpful for testing but technically not needed. (lines 28-29)

## Related Issue
https://github.com/icssc/anteater-api/issues/360

## How Has This Been Tested?
with old vs new sql query

## Types of changes

## Checklist:


laggycomputer (approved):
🥀

PR #366, opened 2026-04-14
feat: Add historical syllabi data endpoint to websoc

### LINKED ISSUE #355: Serve historical syllabi data for courses
There is a syllabus column for course sections in AntAlmanac Scheduler. However, since Professors usually don't provide a syllabus until close to start of a quarter, the column is generally empty when most users are actually planning their next quarter schedule. The data sometimes exists for past the same course taught in past quarters, but to find that data, users have to go through a tedious manual process (find when the professor last taught the course in AA Planner, switch back to Scheduler and go to that specific quarter, and then hope there is a syllabus available).

We want to find a way to expose historical syllabus data through the course endpoint. This might be done by incorporating specific Websoc section syllabus links into the course view similar to how grade distributions are always provided.

aadi-shanker:
## Description

Adds a new GET /v2/rest/websoc/syllabi endpoint that surfaces historical syllabus links from past quarters. Each entry contains the year, quarter, and Canvas URL for the syllabus. Also exposes the same data via GraphQL under syllabi(query: SyllabiQuery!).

## Related Issue

#355 

## Motivation and Context

In AntAlmanac, the syllabus column is almost always empty during planning because professors don't upload syllabi until close to the start of the quarter. However, syllabus URLs for past quarters are already stored in websoc_section.web_url. This change exposes that historical data through a new standalone WebSoc endpoint so students can reference syllabi from previous offerings without manually hunting through old quarters.

## How Has This Been Tested?

Tested locally against GET /v2/rest/websoc/syllabi

## Types of changes

## Checklist:

- [] My code involves a change to the database schema.


apps/api/src/services/websoc-syllabi.ts
```diff
@@ -0,0 +1,45 @@
...
+        quarter: websocSection.quarter,
+        url: websocSection.webURL,
+      })
+      .from(websocSection)
+      .innerJoin(websocCourse, eq(websocSection.courseId, websocCourse.id))
+      .leftJoin(
+        websocSectionToInstructor,
+        eq(websocSectionToInstructor.sectionId, websocSection.id),
+      )
+      .where(and(...conditions))
```
HwijungK: We should return our response reverse chronologically

apps/api/src/schema/websoc-syllabi.ts
```diff
@@ -0,0 +1,25 @@
...
+  instructor: z.string().optional().openapi({
+    description: "Only include syllabi for sections taught by the specified instructor",
+    example: "SHINDLER, M.",
+  }),
+});
+
+export const syllabiSchema = z.object({
+  year: z.string().openapi({ example: "2025" }),
+  quarter: z.enum(terms).openapi({ example: "Fall" }),
+  url: z.string().openapi({
```
HwijungK: would it make sense to also return instructor in our response?

apps/api/src/schema/websoc-syllabi.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from "@hono/zod-openapi";
+import { terms } from "@packages/db/schema";
+import { yearSchema } from "./lib";
+
+export const syllabiQuerySchema = z.object({
+  courseId: z.string().openapi({
+    description: "The course ID to retrieve syllabi for",
+    example: "COMPSCI161",
+  }),
+  year: yearSchema.optional(),
```
HwijungK: yearSchema defaults optional to 2024, but the llast saved syllabi are from 2025. Could we worth overriding the example to 2025 here or in the yearSchema to avoid confusion for people testing with the default values.

HwijungK (changes requested):
I lean towards moving all the code into websoc.ts instead of making a new websoc-syllabi.ts, since everything is contained within websoc side

apps/api/src/schema/websoc-syllabi.ts
aadi-shanker: tried adding it back but it causes duplicates again since each section has a different instructor combo (lecture vs discussion sections). keeping it out for now to maintain clean deduplication you can still filter by instructor via the query param though

apps/api/src/services/websoc.ts
```diff
@@ -443,4 +445,43 @@ export class WebsocService {
...
+    return this.db
+      .select({
+        year: subquery.year,
+        quarter: subquery.quarter,
+        url: subquery.url,
+        instructorNames: sql<string[]>`ARRAY_AGG(${subquery.instructor})`,
+      })
+      .from(subquery)
+      .groupBy(subquery.year, subquery.quarter, subquery.url)
+      .orderBy(desc(subquery.year), desc(subquery.quarter));
```
HwijungK: This acutally ends up sorting in order Fall, Winter ... SummerWk2, so Spring 2026 will be interpreted as later than Fall 2026. Use the sorting strategy used by `getAllTerms`

KevinWu098:
salivating...

HwijungK:
Looking good to me, I'll pass it along to Dante

HwijungK (approved):

apps/api/src/graphql/resolvers/websoc.ts
```diff
@@ -22,5 +23,9 @@ export const websocResolvers = {
         await service.getDepartments(websocDepartmentsQuerySchema.parse(args.query)),
       );
     },
+    syllabi: async (_: unknown, args: { query: unknown }, { db }: GraphQLContext) => {
+      const service = new WebsocService(db);
+      return await service.getSyllabi(syllabiQuerySchema.parse(args.query));
```
laggycomputer: We should Zod validate on the way out both here and on GraphQL (it's already done on GraphQL)

apps/api/src/services/websoc.ts
```diff
@@ -443,4 +445,51 @@ export class WebsocService {
       .where(and(or(...sinceOptions), or(...untilOptions)))
       .orderBy(websocDepartment.deptCode, desc(websocDepartment.year));
   }
+
+  async getSyllabi(input: z.infer<typeof syllabiQuerySchema>) {
+    const conditions = [eq(websocCourse.courseId, input.courseId), gt(websocSection.webURL, "")];
```
laggycomputer: Why `gt`? This is technically correct but has confusing semantics. Use `ne`

laggycomputer (changes requested):

apps/api/src/services/websoc.ts
aadi-shanker: changed to ne good catch, makes the intent clearer I just used gt out of habit.

apps/api/src/graphql/resolvers/websoc.ts
aadi-shanker: fixed added syllabiSchema.array().parse(...) on the way out in the resolver. did you mean "it's already done on REST"? I think the GraphQL resolver was the one missing it

laggycomputer: i am sentient my bad

apps/api/src/services/websoc.ts
```diff
@@ -443,4 +445,51 @@ export class WebsocService {
...
+    }
+    if (input.instructor) {
+      conditions.push(eq(websocSectionToInstructor.instructorName, input.instructor));
+    }
+    const sub = this.db
+      .selectDistinct({
+        year: websocSection.year,
+        quarter: websocSection.quarter,
+        url: websocSection.webURL,
+        instructor: sql<string>`UNNEST(${websocSection.instructors})`.as("instructor"),
```
laggycomputer: Can we avoid doing *both* this unnest and the join? Does this result in O(n^2) times as many rows? Consider using `$1 = ANY(websoc_section.instructors)` to supplant the join, or `$1 = websoc_section_to_instructor.instructor` to supplant the unnest.

laggycomputer (changes requested):

apps/api/src/graphql/resolvers/websoc.ts
aadi-shanker: All g it be like that

apps/api/src/services/websoc.ts
aadi-shanker: fixed removed the join and switched to $1 = ANY(websoc_section.instructors) for the instructor filter as you suggested. only UNNEST remains for the aggregation

laggycomputer (changes requested):
I'm interested in seeing a query planner analysis on whether looking at the instructors array or the table is faster. Otherwise, good.

laggycomputer (approved):

Issue #302, opened 2026-02-09
[websoc] Intermittent loss of GE columns


laggycomputer:
We have experienced multiple instances where the Websoc course table seems to drop GE data.
In particular, we seem to only get complaints (oops!) that GE 4 data is totally empty (i.e. no courses have that column set to `true`, which is certainly not right).
However, this is intermittent and lasts up to several hours.
What this means is that the Websoc scraper is updating this data with an incorrect state, then re-acquiring the correct state and applying that to the database at some point later.

Our first task will be to add more observability stuff to the Websoc scraper so we can see when and why this is happening.

PR #371, opened 2026-04-17
feat(websoc-scraper): added observability during GE scrape

iAmbrosial:
## Description

Adds observability logging to the Websoc scraping pipeline for the GE scrape per term, focusing on highlighting the scrape counts for each GE category within each term. Also adds before and after logging of stored GE counts to allow observation regarding any changes from scrapes over time – especially during the buggy hours. 

This PR also fixed a correctness bug during GE scraping where, previously, GE flags on Websoc courses would never be reset to false if a course lost its GE designation (category specific) between scrapes. This change means stale true flags will not persist indefinitely, even if the likelihood of this is low.

## Related Issue

#302 

## Motivation and Context

We previously experienced issues with the Websoc course table seemingly dropping GE data. These observability changes are an initial step towards figuring out (as much as we can) why GE course data is occasionally dropped. 

## How Has This Been Tested?

Ran Websoc scraper locally and verified GE flag reset logic against the existing DB transactions.

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/websoc-scraper/src/lib.ts
+++ b/apps/data-pipeline/websoc-scraper/src/lib.ts
@@ -732,6 +732,24 @@ type CourseGEUpdate = {
   isGE8?: boolean;
 };
 
+async function getGECountsFromDB(db: ReturnType<typeof database>, term: Term) {
+  return db
+    .select({
+      "GE-1A": sql<number>`count(*) filter (where ${websocCourse.isGE1A})`,
+      "GE-1B": sql<number>`count(*) filter (where ${websocCourse.isGE1B})`,
+      "GE-2": sql<number>`count(*) filter (where ${websocCourse.isGE2})`,
+      "GE-3": sql<number>`count(*) filter (where ${websocCourse.isGE3})`,
+      "GE-4": sql<number>`count(*) filter (where ${websocCourse.isGE4})`,
+      "GE-5A": sql<number>`count(*) filter (where ${websocCourse.isGE5A})`,
+      "GE-5B": sql<number>`count(*) filter (where ${websocCourse.isGE5B})`,
+      "GE-6": sql<number>`count(*) filter (where ${websocCourse.isGE6})`,
+      "GE-7": sql<number>`count(*) filter (where ${websocCourse.isGE7})`,
+      "GE-8": sql<number>`count(*) filter (where ${websocCourse.isGE8})`,
+    })
+    .from(websocCourse)
+    .where(and(eq(websocCourse.year, term.year), eq(websocCourse.quarter, term.quarter)));
+}
+
 async function scrapeGEsForTerm(db: ReturnType<typeof database>, term: Term) {
   const updates = new Map<string, CourseGEUpdate>();
   const outcomes: Record<string, "success" | "empty" | "error"> = {};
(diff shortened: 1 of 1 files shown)


laggycomputer (changes requested):
- Make sure the PR title is Conventional Commits compliant.
- As we discussed in meeting, for each flag, print the count of courses returned and also print the number of rows inserted, so we can see discrepancies.

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -732,18 +732,58 @@ type CourseGEUpdate = {
...
 async function scrapeGEsForTerm(db: ReturnType<typeof database>, term: Term) {
   const updates = new Map<string, CourseGEUpdate>();
+  const outcomes: Record<string, "success" | "empty" | "error"> = {};
+  const scrapedCounts: Record<string, number> = {};
+
+  const before = await getGECountsFromDB(db, term);
+  console.log(`[GE scrape] DB counts BEFORE for ${termToName(term)}:`, before);
+
   for (const ge of geCategories) {
     console.log(`Scraping GE ${ge}`);
```
HwijungK:
```suggestion
    console.log(`Scraping ${ge}`);
```
`ge` is already in form GE-XX

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -772,6 +827,19 @@ async function scrapeGEsForTerm(db: ReturnType<typeof database>, term: Term) {
...
+
+  const after = await getGECountsFromDB(db, term);
+
+  const summary = geCategories.map((ge) => {
+    const stored = after?.[ge] ?? 0;
+    const scraped = scrapedCounts[ge] ?? 0;
+    const outcome = outcomes[ge];
+    const storedLabel = outcome === "error" ? "preserved" : "stored";
+    const diff = outcome === "success" ? scraped - stored : 0;
+    const diffNote = diff > 0 ? `  (${diff} unmatched)` : "";
```
HwijungK: just checking unmatched means that a course used to be a GE but no longer is

HwijungK (changes requested):
Pretty much good i believe.

apps/data-pipeline/websoc-scraper/src/lib.ts
iAmbrosial: Unmatched means that the websoc GE endpoint returned that some course X is a GE, but that course X could not be found in the websocCourse table to be updated – thus causing the GE to not be matched to a course.

laggycomputer: As we're discussing right now, the word "unmatched" is a bit imprecise; we more mean "not found in the database".

laggycomputer (changes requested):

apps/data-pipeline/websoc-scraper/src/lib.ts
```diff
@@ -772,6 +827,19 @@ async function scrapeGEsForTerm(db: ReturnType<typeof database>, term: Term) {
...
+    return `  ${ge}: scraped=${scraped}  ${storedLabel}=${stored}  outcome=${outcome}${diffNote}`;
```
laggycomputer: As we discussed, you should find ways to represent this as a JSON object, which will be nicer to search in Cloudflare.

laggycomputer (changes requested):

laggycomputer (approved):

HwijungK (approved):

PR #376, opened 2026-04-22
fix (study-rooms): fix pruning old time slots

### LINKED ISSUE #375: Study room missing todays rooms
zotmeet found another study room bug
## Expected Behavior
When querying same day rooms, rooms should appear

## Current Behavior
When querying same day rooms, some rooms before a certain time are missing

Image

## Possible Solution
check timezones :C

ZackeryHe:
## Description

was using wrong time

btw probably not worth moving anything else in the time.ts util file bc they are too format specific

## Related Issue
https://github.com/icssc/anteater-api/issues/375

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/study-location-scraper/src/time.ts
+++ b/apps/data-pipeline/study-location-scraper/src/time.ts
@@ -1,0 +1,25 @@
+export function getLaWallClockAsUtcDate(): Date {
+  const parts = new Intl.DateTimeFormat("en-CA", {
+    timeZone: "America/Los_Angeles",
+    year: "numeric",
+    month: "2-digit",
+    day: "2-digit",
+    hour: "2-digit",
+    minute: "2-digit",
+    second: "2-digit",
+    hour12: false,
+  }).formatToParts(new Date());
+
+  const values = Object.fromEntries(parts.map((part) => [part.type, part.value]));
+
+  return new Date(
+    Date.UTC(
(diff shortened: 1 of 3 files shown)


apps/data-pipeline/study-location-scraper/src/time.ts
```diff
@@ -0,0 +1,25 @@
+export function getLaWallClockAsUtcDate(): Date {
+  const parts = new Intl.DateTimeFormat("en-CA", {
```
laggycomputer: This is Canadian English ("Californian English" does not exist, no matter how different we think valley people sound)! Was that intended?

apps/data-pipeline/study-location-scraper/src/time.ts
```diff
@@ -0,0 +1,25 @@
+export function getLaWallClockAsUtcDate(): Date {
```
laggycomputer: This name is a little strange. Why not something like `losAngelesNowNaive`?

laggycomputer (changes requested):
I still think the best thing to do is just to add a timezone to the database columns, instead of futzing with JS time APIs.

Also see naming on the other PR.

ZackeryHe:
we could do changing to timestamptz in another issue. and itll probably make adding ALP support easier

laggycomputer (approved):

Issue #243, opened 2025-10-24
Library Traffic Data Averaging Using Historical Data


estao1:
## Expected Behavior

The API should provide averaged traffic data based on historical patterns stored in the `library_traffic_history` table. Users should be able to query typical occupancy levels at different times to help plan their study sessions.

**Examples:**
- Average traffic for Langson Library 4th Floor on Wednesdays at 2pm
- Average afternoon traffic for Gateway Study Center across all weekdays
- Daily average traffic for Science Library during finals week

## Current Behavior

The library traffic feature currently:
- Scrapes traffic data every 15 minutes when libraries are open ("active" status)
- Scrapes traffic data every 60 minutes when libraries are closed ("idle" status)  
- Stores current snapshot in `library_traffic` table
- Accumulates historical data in `library_traffic_history` table with timestamps

The API only returns the most recent snapshot via `/library-traffic` endpoint. Historical data is being collected but not exposed or aggregated for analysis.

## Possible Solution

Add new API endpoints that leverage the historical data to provide averaged traffic patterns:

### Averaging Approaches/Features to Consider

1. **Hourly Averages**: Average traffic per hour of the day (e.g. 2pm on Mondays)
2. **Daily Averages**: Average traffic per day of the week (e.g. Tuesdays)
3. **Time-of-Day Averages**: Average traffic across different time blocks (e.g. morning, afternoon, evening)
4. **Day-Specific Averages**: Average traffic on a specific Calendar day (e.g. Wednesday of finals week)
5. **Custom Time Ranges**: Allow users to specify date ranges for custom averaging

PR #379, opened 2026-04-22
fix(library-traffic): reduce library traffic scrape load

### LINKED ISSUE #268: Reduce library traffic scrape frequency
Library traffic is currently scraping around ~450,000 times a year, which is too much. We need to cut this number down, along with only keeping traffic data for up to a year.

estao1:
## Description

Reduces library traffic scraper load by halving the scrape frequency and pruning stale history records.

- Cron schedule changed from `*/15 * * * *` to `*/30 * * * *` — scraper now runs every 30 minutes instead of 15
- On each scrape run, history records older than 13 months are deleted before new data is inserted — this keeps one full year of data plus a one-quarter buffer for year-over-year comparisons (e.g. Fall 2025 finals vs Fall 2026 finals) without unbounded table growth

## Related Issue

#268 

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/library-traffic-scraper/src/lib.ts
+++ b/apps/data-pipeline/library-traffic-scraper/src/lib.ts
@@ -206,7 +206,7 @@ export async function doScrape(db: ReturnType<typeof database>) {
   console.log("Starting library traffic scrape.");
 
   const cutoff = new Date();
-  cutoff.setMonth(cutoff.getMonth() - 13);
+  cutoff.setMonth(cutoff.getMonth() - 15);
   await db.delete(libraryTrafficHistory).where(lt(libraryTrafficHistory.timestamp, cutoff));
 
   const locationMeta = await collectLocationMeta();
--- a/apps/data-pipeline/library-traffic-scraper/src/lib.ts
+++ b/apps/data-pipeline/library-traffic-scraper/src/lib.ts
@@ -1,4 +1,5 @@
 import type { database } from "@packages/db";
+import { lt } from "@packages/db/drizzle";
 import { libraryTraffic, libraryTrafficHistory } from "@packages/db/schema";
 import { conflictUpdateSetAllCols } from "@packages/db/utils";
 import { load } from "cheerio";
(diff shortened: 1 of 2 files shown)


KevinWu098:
Question! Why do we not store historical data? Question mark?

estao1:
> Question! Why do we not store historical data? Question mark?

we do, just need to reduce it (currently around ~450,000 rows/year). endpoint to query historical data is in the works (#243)

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -203,6 +204,11 @@ const libraryCodeMap = {
 
 export async function doScrape(db: ReturnType<typeof database>) {
   console.log("Starting library traffic scrape.");
+
+  const cutoff = new Date();
+  cutoff.setMonth(cutoff.getMonth() - 13);
```
laggycomputer: One month isn't enough to guarantee that data is included back to the start of the current quarter in the previous year. That would require a year plus 11 weeks, or just under 15 months. Do you want to change this?

laggycomputer (changes requested):

apps/data-pipeline/library-traffic-scraper/src/lib.ts
```diff
@@ -203,6 +204,11 @@ const libraryCodeMap = {
 
 export async function doScrape(db: ReturnType<typeof database>) {
   console.log("Starting library traffic scrape.");
+
+  const cutoff = new Date();
+  cutoff.setMonth(cutoff.getMonth() - 15);
+  await db.delete(libraryTrafficHistory).where(lt(libraryTrafficHistory.timestamp, cutoff));
```
laggycomputer: Can we do this after everything is inserted? Say we had a crash loop: I'd rather crash and retain extra data so we always have a year of something (even if it's a bit stale) than start wiping past data without getting new data.

laggycomputer (changes requested):
one more thing

laggycomputer (approved):

PR #385, opened 2026-04-30
feat: Low/No-Cost Course Materials

matt-franklin225:
## Description
This creates an endpoint for low and no-cost course materials (particularly books), which is tracked by UCI Libraries and publicized on [this page](https://www.lib.uci.edu/affordable-initiatives/course-materials). We now import this data (which is provided in the form of a spreadsheet provided by UCI Libraries) through `apps/data-pipeline/course_materials-importer` and store it in a new table called `course_material`. Some details relevant to the textbook, such as instructor, course number/department, etc. are retrieved through `websoc_course` and `websoc_section`. This data is now accessible through a new endpoint (called "Filter course materials") under the Course Materials tag. Through this endpoint, users can choose various attributes (year, quarter, section code, etc.) to filter the data.

## How Has This Been Tested?
These changes have been tested locally. The importing was tested comparing the data imported into the database with that in the spreadsheet. The endpoint was tested through a variety of queries with different filters applied.

image

image

## Types of changes

## Checklist:


apps/data-pipeline/course-materials-importer/src/index.ts
```diff
@@ -0,0 +1,110 @@
...
+      continue;
+    }
+
+    values.push({
+      sectionId: section.id,
+      title: getFromMapOrThrow(entry, "Title"),
+      author: entry.get("Author"),
+      edition: entry.get("Edition"),
+      isbn: entry.get("ISBN"),
+      format: (entry.get("Format") || null) as any,
```
HwijungK: Is there a reason we can't provide the actual type `TextbookFormat`?

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,101 @@
...
+function buildQuery(input: CourseMaterialsServiceInput) {
+  const conditions = [];
+  if (input.year) {
+    conditions.push(eq(websocCourse.year, input.year));
+  }
+  if (input.quarter) {
+    if (input.quarter === "Summer") {
+      conditions.push(inArray(websocCourse.quarter, ["Summer1", "Summer2", "Summer10wk"] as any[]));
+    } else {
+      conditions.push(eq(websocCourse.quarter, input.quarter as any));
```
HwijungK: are these `as any` necessary? Doesn't look like it to me first glance but i may be missing something.

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,101 @@
...
+type CourseMaterialsServiceInput = z.infer<typeof courseMaterialsQuerySchema>;
+
+function buildQuery(input: CourseMaterialsServiceInput) {
+  const conditions = [];
+  if (input.year) {
+    conditions.push(eq(websocCourse.year, input.year));
+  }
+  if (input.quarter) {
+    if (input.quarter === "Summer") {
+      conditions.push(inArray(websocCourse.quarter, ["Summer1", "Summer2", "Summer10wk"] as any[]));
```
HwijungK: same as below

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+      conditions.push(eq(websocCourse.quarter, input.quarter as any));
+    }
+  }
+  if (input.instructor) {
+    conditions.push(eq(websocSectionToInstructor.instructorName, input.instructor));
```
HwijungK: would like to see this using iLike as we do with `author` and `title`

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,101 @@
...
+      .where(buildQuery(input));
+
+    return rows
+      .reduce((acc, row) => {
+        if (row && row.materialId && !acc.has(row.materialId)) {
+          const displayQuarter = row.quarter.startsWith("Summer") ? "Summer" : row.quarter;
+          acc.set(row.materialId, {
+            ...row,
+            quarter: displayQuarter as any,
+            sectionCode: row.sectionCode ? row.sectionCode.toString(10).padStart(5, "0") : "00000",
```
HwijungK: What's the reasoning behind making a non-existant sectionCode "00000" instead of setting the sectionCode as nullable? I understand that "00000" can never be a valid course but it feels wrong to me.

HwijungK (changes requested):
Everything seems sane to me and the endpoint looks like it works well. There's a lot of casts to type 'any' which I can't see the reason for. There's also some comments that are more questions then fixes so let me know your thoughts on those.

matt-franklin225:
> Everything seems sane to me and the endpoint looks like it works well. There's a lot of casts to type 'any' which I can't see the reason for. There's also some comments that are more questions then fixes so let me know your thoughts on those.

Resolved all of those comments now, I also slightly adjusted a couple things in the schema (added indexing and removed some unused code). I don't think the changes should significantly impact everything, but I would recommend re-running the migration.

apps/api/src/schema/course-materials.ts
```diff
@@ -0,0 +1,56 @@
...
+  quarter: z.enum(materialTerms, { error: (_issue) => "Invalid quarter provided" }).optional(),
+  department: z.string().optional().openapi({
+    description: "Only include materials from courses offered by the specified department code",
+    example: "I&C SCI",
+  }),
+  courseNumber: z.string().optional().openapi({
+    description: "Only include materials from courses exactly matching the specified course number",
+    example: "45C",
+  }),
+  sectionCode: z
```
HwijungK: We should consider turning this into a comma seperated list. We already have the code for that in schema/lib/course.ts

HwijungK (changes requested):

HwijungK:
Looks good to me. Will link Dante!

HwijungK (approved):

apps/api/src/graphql/resolvers/course-materials.ts
```diff
@@ -0,0 +1,12 @@
+import type { GraphQLContext } from "$graphql/graphql-context";
+import { courseMaterialsQuerySchema } from "$schema";
+import { CourseMaterialsService } from "$services";
+
+export const courseMaterialsResolvers = {
+  Query: {
+    samplePrograms: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
```
laggycomputer: Wrong name...

apps/api/src/graphql/schema/course-materials.ts
```diff
@@ -0,0 +1,39 @@
...
+    Physical
+    Electronic
+    Both
+    OER
+}
+
+enum MaterialRequirement {
+    Required
+    Recommended
+    Go to Class First
```
laggycomputer: Make this PascalCase as discussed

apps/api/src/schema/course-materials.ts
```diff
@@ -0,0 +1,55 @@
...
+  courseNumeric: z.number(),
+  instructors: z.string().array(),
+  author: z.string().nullable(),
+  title: z.string(),
+  edition: z.string().nullable(),
+  format: z.enum(textbookFormats),
+  requirement: z.enum(materialRequirements).nullable(),
+  isbn: z.string().nullable(),
+  mmsId: z.string().nullable(),
+  link: z.string().nullable(),
```
laggycomputer: Document + examples for these fields, especially some of the highlighted ones

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,102 @@
...
+      .innerJoin(courseMaterial, eq(courseMaterial.sectionId, websocSection.id))
+      .leftJoin(
+        websocSectionToInstructor,
+        eq(websocSection.id, websocSectionToInstructor.sectionId),
+      )
+      .where(buildQuery(input));
+
+    return rows
+      .reduce((acc, row) => {
+        if (row && row.materialId && !acc.has(row.materialId)) {
```
laggycomputer: First two conditions are reducible to `row?.foo`

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,102 @@
...
+          const displayQuarter = row.quarter.startsWith("Summer") ? "Summer" : row.quarter;
+          acc.set(row.materialId, {
+            ...row,
+            quarter: displayQuarter as any,
+            sectionCode: row.sectionCode.toString(10).padStart(5, "0"),
+          });
+        }
+        return acc;
+      }, new Map<string, any>())
```
laggycomputer: Use something more strict than `any`

packages/db/src/schema.ts
```diff
@@ -868,6 +878,23 @@ export const libraryTrafficHistory = pgTable(
   (table) => [uniqueIndex().on(table.locationId, table.timestamp)],
 );
 
+export const courseMaterial = pgTable(
+  "course_material",
+  {
+    id: uuid("id").primaryKey().defaultRandom(),
+    sectionId: uuid("section_id").references(() => websocSection.id),
```
laggycomputer: Why not `NOT NULL`?

laggycomputer (changes requested):

apps/api/src/graphql/schema/course-materials.ts
```diff
@@ -16,7 +16,7 @@ enum TextbookFormat {
 enum MaterialRequirement {
     Required
     Recommended
-    Go to Class First
+    GoToClassFirst
```
laggycomputer, line 19: database insert and zod schema are now probably wrong

laggycomputer (changes requested):

apps/api/src/rest/routes/course-materials.ts
```diff
@@ -0,0 +1,48 @@
...
+  response422,
+  response500,
+} from "$schema";
+import { CourseMaterialsService } from "$services";
+
+const courseMaterialsRouter = new OpenAPIHono<{ Bindings: Env }>({ defaultHook });
+
+const filterCourseMaterialsRoute = createRoute({
+  summary: "Filter course materials",
+  operationId: "filterCourseMaterials",
```
laggycomputer: These should be about "getting course materials". Also change GraphQL accordingly.

laggycomputer (changes requested):
missing `extend type Query` to create new graphql resource

apps/api/src/schema/course-materials.ts
matt-franklin225: Added documentation/examples for the ones highlighted, lmk if you want the others as well

apps/api/src/schema/course-materials.ts
```diff
@@ -33,7 +39,10 @@ export const courseMaterialsQuerySchema = z.object({
     example: "ALGORITHM DESIGN",
   }),
   format: z.enum(textbookFormats).optional(),
-  requirement: z.enum(materialRequirements).optional(),
+  requirement: z.preprocess(
+    (val) => (typeof val === "string" ? (requirementMap[val] ?? val) : val),
+    z.enum(materialRequirements).optional(),
+  ),
```
laggycomputer: Zod is not a good place for logic like this; it can be overlooked while auditing. Do this in endpoint logic.

apps/api/src/services/course-materials.ts
laggycomputer: If we're going to require `materialId` to be truthy, and it's a row, why not apply a SQL `WHERE`?

apps/api/src/graphql/schema/course-materials.ts
```diff
@@ -19,6 +19,19 @@ enum MaterialRequirement {
...
+    year: String
+    quarter: MaterialTerm
+    department: String
+    courseNumber: String
+    sectionCode: String
+    instructors: String
+    author: String
+    title: String
+    format: TextbookFormat
+    requirement: MaterialRequirement
```
laggycomputer: I don't think some of these are nullable...

laggycomputer (changes requested):

apps/api/src/services/course-materials.ts
matt-franklin225: Check to see if new Map<> is correct

apps/api/src/schema/course-materials.ts
matt-franklin225: Would that be in the graphql resolver?

laggycomputer: In the service file so that it's shared between REST and GraphQL.

matt-franklin225:
Due to inconsistencies with the "Go to class first" naming I just adjusted the code so that that info is stored as "GoToClassFirst" with no spaces in the db and retrieved in the same way, consistent naming across rest/graphql/db.

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export class CourseMaterialsService {
+  constructor(private readonly db: ReturnType<typeof database>) {}
+
+  async getCourseMaterials(input: CourseMaterialsServiceInput) {
+    const rows = await this.db
+      .select({
+        materialId: courseMaterial.id,
+        id: websocSection.id,
```
laggycomputer: Why select this?

apps/api/src/services/course-materials.ts
```diff
@@ -0,0 +1,103 @@
...
+        year: websocSection.year,
+        quarter: websocSection.quarter,
+        sectionCode: websocSection.sectionCode,
+        department: websocCourse.deptCode,
+        courseNumber: websocCourse.courseNumber,
+        courseNumeric: websocCourse.courseNumeric,
+        instructors: websocSection.instructors,
```
laggycomputer, line 68: Did you intend to include TAs here? If not, use `websoc_section_to_instructor`

apps/api/src/graphql/resolvers/course-materials.ts
```diff
@@ -2,11 +2,22 @@ import type { GraphQLContext } from "$graphql/graphql-context";
...
+const requirementMap: Record<string, string> = {
+  Required: "Required",
+  Recommended: "Recommended",
+  GoToClassFirst: "Go to Class First",
+};
+
 export const courseMaterialsResolvers = {
   Query: {
-    getCourseMaterials: async (_: unknown, args: { query?: unknown }, { db }: GraphQLContext) => {
+    getCourseMaterials: async (_: unknown, args: { query?: any }, { db }: GraphQLContext) => {
```
HwijungK: Do we still need query as `any` since we've fixed it to zod parse first?

HwijungK (changes requested):

laggycomputer (changes requested):

apps/api/src/graphql/resolvers/course-materials.ts
laggycomputer: It technically is `unknown` at this point since it's actually not been parsed yet, IIRC.

apps/api/src/services/course-materials.ts
matt-franklin225: Including TAs was intentional, although I can switch over if that info is considered unimportant

laggycomputer: Eh, arguably none of this data is necessary. Including them is fine.

laggycomputer (changes requested):
- You should probably require at least one of the filters, kind of like what WebSoc does. I think there could be multiple allowable minimum combinations. Section code alone should be acceptable. So should department and course number. Maybe some of the other fields too.
- If any of the other fields are allowed as a minimum filter set, particularly those which come from the new table and not the WebSoc section table, add indices on them. For example, if you make edition a valid filter set (meaning you're allowed to specify no other filters), you should have an index on the edition column. Don't actually do that, because filtering for edition and nothing else is silly.

matt-franklin225:
> * You should probably require at least one of the filters, kind of like what WebSoc does. I think there could be multiple allowable minimum combinations. Section code alone should be acceptable. So should department and course number. Maybe some of the other fields too.
> 
>     * If any of the other fields are allowed as a minimum filter set, particularly those which come from the new table and not the WebSoc section table, add indices on them. For example, if you make edition a valid filter set (meaning you're allowed to specify no other filters), you should have an index on the edition column. Don't actually do that, because filtering for edition and nothing else is silly.

Search now requires one of the following: sectionCode/year/quarter or department/courseNumber. The indexes for these are on other tables, so I haven't added any new ones. Should I add any more options/alter these, and is there any more indexing to do?

apps/api/src/graphql/schema/course-materials.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+extend type Query {
+    getCourseMaterials(query: GetCourseMaterialsQuery): [CourseMaterial!]!
```
laggycomputer: Rename this and all related stuff for GraphQL to exclude "get" because that's already implied

apps/api/src/schema/course-materials.ts
```diff
@@ -0,0 +1,91 @@
...
+    }),
+    format: z
+      .enum(textbookFormats, { error: (_issue) => "Invalid textbook format provided" })
+      .optional(),
+    requirement: z
+      .enum(materialRequirements, { error: (_issue) => "Invalid requirement provided" })
+      .optional(),
+  })
+  .refine((x) => (x.department && x.courseNumber) || (x.sectionCode && x.year && x.quarter), {
+    message: "Must provide department and course number; or section code and year/quarter",
```
laggycomputer: erm actually you don't get to use a semicolon unless you have comma lists

i also don't care and this is stupid

laggycomputer (changes requested):
can you open an issue to investigate DISTINCT ON in the service layer query

it might be faster but we don't have time so we can ship without

apps/api/src/graphql/schema/course-materials.ts
matt-franklin225: Do I do this for rest and the function in services as well?

laggycomputer: no

laggycomputer (approved):

apps/api/src/schema/course-materials.ts
```diff
@@ -2,39 +2,47 @@ import { z } from "@hono/zod-openapi";
...
+        "Only include materials from the specified author (case-insensitive, last name only)",
+      example: "KLEINBERG",
+    }),
+    title: z.string().optional().openapi({
+      description:
+        "Only include materials whose title contains the specified string (case-insensitive)",
+      example: "ALGORITHM DESIGN",
+    }),
+    format: z
+      .enum(textbookFormats, { error: (_issue) => "Invalid textbook format provided" })
```
HwijungK: do the errors have to be an arrow function sinse we don't use `_issue`

apps/api/src/schema/course-materials.ts
```diff
@@ -2,39 +2,47 @@ import { z } from "@hono/zod-openapi";
...
+      .optional(),
+    requirement: z
+      .enum(materialRequirements, { error: (_issue) => "Invalid requirement provided" })
```
HwijungK: here too

HwijungK (changes requested):

matt-franklin225:
Migration number updated to 31

HwijungK (approved):
Is looking good. Sinse this change involves a migration hold off on merging for now though

PR #386, opened 2026-05-01
fix: Redirected Major Requirements to Other Block

### LINKED ISSUE #321: DegreeWorks: Redirected Major Requirements
For English (BA-345) and History (BA-429), the major requirement when not taking any  specialization is empty.
This is because the major requirements on degreeworks doesn't exist in the "MAJOR" block, but in an "OTHER" block, with the major block having a rule that points to the "OTHER" block. This is similar to how major blocks reference specialization blocks (if one is being taken). Currently the `AuditParser` ignore "block" rules, so these requirements are missed.

## Possible Solution

The main problem is that a call to `parseBlock` only takes in one block, so even when we find a "block" rule type, we have no reference to that other block. We could call parseBlock on "OTHER" block seperately in `Scraper.ts` when we parse the major blocks (similar to how we also parse schoolRequirements at the same time). We should first research if all "OTHER" blocks are extensions of a major requirement, or if they can mean something else.

limlindsayj:
## Description
If the MAJOR block's requirements points to an OTHER block, that major's OTHER that contains the requirements get appended to the Major's json output. This happens in DegreeworksClient.ts. This Major json output gets put into Scraper.ts, which calls AuditParser.ts on that Major json output to parse it. If it contains "otherBlock", then it uses "otherBlock"'s rule array instead of the MAJOR block's rule array, since the requirements are stored there. 

## Related Issue
#321

## Motivation and Context
Majors Classics, History, and English store their requirements in an OTHER block, atypical of how the other majors store their requirements directly in the MAJOR block. This causes the requirements of these majors to be missed when parsing the information for these majors. 

## How Has This Been Tested?
The degreeworks scraper was run and the database is now updated to have requirements for History and English. However, the Classics major is split into specializations, so the requirements are stored there. 

## Types of changes

## Checklist:


apps/data-pipeline/degreeworks-scraper/src/components/AuditParser.ts
```diff
@@ -21,17 +19,19 @@ export class AuditParser {
...
+  parseBlock = async (
+    blockId: string,
+    block: Block & { otherBlock?: Block },
+  ): Promise<DegreeWorksProgram> => ({
     ...this.parseBlockId(blockId),
     name: block.title,
-    requirements: await this.ruleArrayToRequirements(block.ruleArray),
+    requirements: await this.ruleArrayToRequirements(
+      block.otherBlock?.ruleArray ?? block.ruleArray,
+    ),
```
laggycomputer: As we discussed in meeting, we should combine the requirements from the MAJOR and OTHER block, instead of using OTHER alone.

apps/data-pipeline/degreeworks-scraper/src/components/DegreeworksClient.ts
```diff
@@ -145,17 +134,30 @@ export class DegreeworksClient {
...
-        (x) => x.requirementType === "MAJOR" && x.requirementValue === majorCode,
-      ),
+      major: (isOtherBlock
+        ? {
+            ...major,
+            otherBlock: json.blockArray.find(
+              (x) => x.requirementType === "OTHER" && x.title.startsWith("Major in"),
+            ),
+          }
+        : major) as (Block & { otherBlock?: Block }) | undefined,
```
laggycomputer: There should be a comment here describing the premise of the issue, possibly listing a few examples of majors where this happens, and explaining the course of action we take.

apps/data-pipeline/degreeworks-scraper/src/components/DegreeworksClient.ts
```diff
@@ -125,7 +114,7 @@ export class DegreeworksClient {
   ): Promise<
     | {
         college?: Block;
-        major?: Block;
+        major?: Block & { otherBlock?: Block };
```
laggycomputer: I'm uncomfortable with nesting the "other" block *inside* the major block. Can they be placed side by side or with the "other" block as a sibling field?

laggycomputer (changes requested):

laggycomputer (changes requested):
Code looks good, pending merge conflicts.

laggycomputer (changes requested):
Can we revert the other unrelated changes and fix merge conflicts?

apps/data-pipeline/degreeworks-scraper/src/components/DegreeworksClient.ts
```diff
@@ -149,13 +150,21 @@ export class DegreeworksClient {
 
     const json = await this.parseResponse(res, dwAuditOKResponseSchema, "audit");
     if (!json) return undefined;
-
+    if ("error" in json) {
+      return undefined;
+    }
```
laggycomputer: What's the reason for this? We already assert against the error variant in Zod on line 151.

laggycomputer (changes requested):

laggycomputer (approved):
One unnecessary format change but that's fine.

Good work!

HwijungK (approved):

PR #389, opened 2026-05-02
refactor: Helper to Sort Websoc Terms

### LINKED ISSUE #382: Move websoc term sorting to SQL ORDER BY with shared helper
## Expected Behavior

WebSoc term sorting (year DESC, then quarter DESC by academic order: Fall -> Summer2 -> Summer1 -> Spring -> Winter) can be handled in SQL using ORDER BY with a shared helper in the db package, rather than in JavaScript after the query. (Compute effeciency is better)

## Current Behavior

Several places in the codebase sort WebSoc terms in JavaScript using a termOrder map and .sort() + .reverse() after fetching results. SQL compute is cheaper than JS compute and this pattern should be consolidated.

## Possible Solution

Create a shared term ordering helper in @packages/db that can be reused across getAllTerms, getSyllabi, and any other place that sorts by WebSoc term. Use ORDER BY year DESC, CASE quarter WHEN 'Fall' THEN 5 ... END DESC in the query instead of JS-side sorting.

## Context
Identified during review of PR #366. @laggycomputer suggested moving ordering to SQL and creating a helper in the db package so the pattern is consistent across all websoc term sorting in the codebase.

JasonNguyen067:
## Description

Created a helper that sorts by websocTerm and is reused in getAllTerms and getSyllabi. 

## Related Issue

#382 

## Motivation and Context

Sorting via SQL and not JS is more efficient in terms of resources. Using a helper also fixes redundant code and overall simplifies it.

## How Has This Been Tested?

Endpoints returned true.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/services/websoc.ts
+++ b/apps/api/src/services/websoc.ts
@@ -27,7 +27,7 @@ import {
   websocSectionMeetingToLocation,
   websocSectionToInstructor,
 } from "@packages/db/schema";
-import { isFalse, isTrue } from "@packages/db/utils";
+import { isFalse, isTrue, websocTermSortOrder } from "@packages/db/utils";
 import { negativeAsNull } from "@packages/stdlib";
 import type { z } from "zod";
 import type {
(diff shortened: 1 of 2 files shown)


laggycomputer (changes requested):
Thanks for the PR! Two things:
- Unlike AntAlmanac Planner, we enforce [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages, which means your PR title must conform.
- See if you can remove more usages of `termOrder`. I can already see one additional opportunity to do so and there may be more.

laggycomputer (changes requested):
Good that all sorting is now achieved through SQL.

Now:
- You inserted the new function between an existing name and its documentation, so you should move your new function somewhere above or below those.
- Exporting lambdas with no captures is a bit of an antipattern, so make that into a function.
- You should use the type system to decrease the likelihood that a column is improperly passed to the function. Try requiring that the column have type `PgColumn<T extends ColumnBaseConfig<"string", "PgEnumColumn"> & { enumValues: (typeof terms)[number] }>>`

laggycomputer (approved):
I'm satisfied with this. If you'd like to make the types more correct, guarantee that the `string` is one of the valid quarter variants.

HwijungK (changes requested):
Looking good to me. 

Let's convert the input `string` type to match valid terms like Dante suggested. and also make the PR title a little more descriptive. something like `refactor: Helper to Sort Websoc Terms` would be good.

HwijungK (approved):
Looks good to me. Thanks for all your effort; we'll get this PR merged shortly

JasonNguyen067:
> Looks good to me. Thanks for all your effort; we'll get this PR merged shortly

Thanks, appreciate the code review

Issue #186, opened 2025-07-25
I&C SCI H197 misbehaves in search


laggycomputer:
The search string "H197" has been reported to fail to bring up I&C SCI H197, Honors Seminar.
The course is known in our system.
The search string "I&C SCI 197" works as might be expected, as does "197".

Seems our search problems are not behind us.

laggycomputer:
it works now

PR #390, opened 2026-05-06
fix(search): course search query adjustment fixes

ParzivalPerhaps:
## Description

The primary change here is that we no longer split the letter from the course code as we were previously, **i.e.**
`ICS H32` -> `ICS H 32` this seemed to be disrupting Postgres FTS as after this step was removed searches like H197 and H32 seemed to function perfectly fine. Perhaps there is some tribal knowledge that has been lost related to this but search seems to perform extremely similarly with the obvious exception that special courses like H32 and H197 pop up with more ergonomic query patterns.

I also added a small replacement on course searches that converts ICS to I&CSCI since most people colloquially refer to ICS courses as ICS ##, this could also be applied to other departments if there are other relevant examples.

I tested this with a local environment of AntAlmanac Planner.

## Related Issue
#186 

## Types of changes

### DIFF
--- a/apps/api/src/services/search.ts
+++ b/apps/api/src/services/search.ts
@@ -42,7 +42,7 @@ function toQuery(query: string, resultType?: "instructor" | "course") {
   if (resultType === "course")
     for (const code in DEPT_TO_ALIAS)
       normalizedQuery = normalizedQuery.replaceAll(
-        new RegExp(`(^|s)${DEPT_TO_ALIAS[code as DeptCode].toLowerCase()}`, "gi"),
+        new RegExp(`(^|\\s)${DEPT_TO_ALIAS[code as DeptCode].toLowerCase()}`, "gi"),
         `${code.toLowerCase()} `,
       );
 
--- a/apps/api/src/services/search.ts
+++ b/apps/api/src/services/search.ts
@@ -42,7 +42,7 @@ function toQuery(query: string, resultType?: "instructor" | "course") {
   if (resultType && resultType === "course")
     for (const code in DEPT_TO_ALIAS)
       normalizedQuery = normalizedQuery.replaceAll(
-        `${DEPT_TO_ALIAS[code as DeptCode].toLowerCase()} `,
+        new RegExp(`(^|s)${DEPT_TO_ALIAS[code as DeptCode].toLowerCase()}`, "gi"),
         `${code.toLowerCase()} `,
       );
 
--- a/apps/api/src/services/search.ts
+++ b/apps/api/src/services/search.ts
@@ -31,14 +31,17 @@ const INSTRUCTORS_WEIGHTS = sql`(
   SETWEIGHT(TO_TSVECTOR('english', COALESCE(${instructor.title}, '')), 'B')
   )`;
 
-function toQuery(query: string) {
-  const normalizedQuery = query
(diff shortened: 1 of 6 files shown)


apps/api/src/services/search.ts
```diff
@@ -31,22 +31,21 @@ const INSTRUCTORS_WEIGHTS = sql`(
...
     .split(" ")
     .map((x) => x.replace(/^\\-/, "-"))
     .join(" ");
+
+  if (resultType && resultType === "course")
+    for (const code in DEPT_TO_ALIAS)
+      normalizedQuery = normalizedQuery.replaceAll(
+        `${DEPT_TO_ALIAS[code as DeptCode].toLowerCase()} `,
+        `${code.toLowerCase()} `,
+      );
```
laggycomputer, line 47: This will replace "statics " with "stati&c sci " (lol), so you will need to be stricter.

laggycomputer (changes requested):

apps/api/src/services/search.ts
```diff
@@ -31,22 +31,21 @@ const INSTRUCTORS_WEIGHTS = sql`(
...
     .replaceAll(/ {2,}/g, " ")
     .replaceAll(/-/g, "\\-")
     .split(" ")
     .map((x) => x.replace(/^\\-/, "-"))
     .join(" ");
+
+  if (resultType && resultType === "course")
+    for (const code in DEPT_TO_ALIAS)
+      normalizedQuery = normalizedQuery.replaceAll(
+        new RegExp(`(^|s)${DEPT_TO_ALIAS[code as DeptCode].toLowerCase()}`, "gi"),
```
laggycomputer: that is a single literal s not `\\s+` which would match any whitespace(s)

apps/api/src/services/search.ts
```diff
@@ -31,22 +31,21 @@ const INSTRUCTORS_WEIGHTS = sql`(
...
-  const normalizedQuery = splitAtLastNumber(query)
+function toQuery(query: string, resultType?: "instructor" | "course") {
+  let normalizedQuery = query
     .replaceAll(/ {2,}/g, " ")
     .replaceAll(/-/g, "\\-")
     .split(" ")
     .map((x) => x.replace(/^\\-/, "-"))
     .join(" ");
+
+  if (resultType && resultType === "course")
```
laggycomputer: What about a missing `resultType` which would indicate a mixed search type including both instructors and courses? Are you more concerned that this never apply to instructors or that it always apply to courses? In the latter case, allow a missing `resultType` here as well.

laggycomputer (changes requested):

apps/api/src/services/search.ts
ParzivalPerhaps: oops I deleted a character before committing 💀

laggycomputer: reasonable

laggycomputer (changes requested):
you didn't actually fix the escape

ParzivalPerhaps:
> you didn't actually fix the escape

I'm getting ragebaited by vs code

laggycomputer (changes requested):
this is going to replace "WRITING" with "WRITINGITING"

laggycomputer (approved):
laggo impressed

HwijungK (approved):

PR #391, opened 2026-05-06
feat (dining): serve past events

### LINKED ISSUE #380: Dining: serve past events
(@EightBitByte)

We'd like to serve past events, instead of only looking at present and future.
It's been tentatively agreed that `/dining/events` ought to be expanded to have optional date bound parameters.
Perhaps the default with these parameters excluded would be the current behavior (i.e. events starting not after the current time)?

ZackeryHe:
## Related Issue
https://github.com/icssc/anteater-api/issues/380

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/graphql/schema/dining.ts
+++ b/apps/api/src/graphql/schema/dining.ts
@@ -100,6 +100,8 @@ type RestaurantToday @cacheControl(maxAge: 3600) {
 
 input DiningEventsQuery {
   restaurantId: RestaurantId
+  startDate: String
+  endDate: String
 }
 
 input RestaurantsQuery {
--- a/apps/api/src/graphql/schema/dining.ts
+++ b/apps/api/src/graphql/schema/dining.ts
@@ -133,8 +133,8 @@ type Schedule @cacheControl(maxAge: 3600) {
 
 input DiningEventsQuery {
   restaurantId: RestaurantId
-  startDate: String
-  endDate: String
+  after: String
+  before: String
 }
 
 input RestaurantsQuery {
--- a/apps/api/src/rest/routes/dining.ts
+++ b/apps/api/src/rest/routes/dining.ts
@@ -32,7 +32,7 @@ const eventsRoute = createRoute({
   path: "/events",
   request: { query: diningEventsQuerySchema },
   description:
-    "Retrieves all ongoing and upcoming dining events at the current timestamp. Events with null end time are served for 2 weeks after their removal from the underlying data source,",
+    "Retrieves all ongoing and upcoming dining events at the current timestamp. Events with null end time are served for 2 weeks after their removal from the underlying data source. Use startDate and/or endDate to include past events.",
   responses: {
     200: response200(diningEventsResponseSchema),
     422: response422(),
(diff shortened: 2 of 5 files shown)


apps/api/src/services/dining.ts
```diff
@@ -28,14 +40,25 @@ type DiningEventQuery = z.infer<typeof diningEventsQuerySchema>;
...
-    const conds = [
-      or(
-        gte(diningEvent.end, sql`NOW()`),
-        and(isNull(diningEvent.end), gte(diningEvent.updatedAt, sql`NOW() - INTERVAL '2 WEEKS'`)),
-      ),
-    ];
+  async getEvents(input: DiningEventQuery) {
+    const conds = [];
+
+    if (input.startDate || input.endDate) {
```
HwijungK: unnecessary enclosing if statement? can we just use if, else if, else

HwijungK (commented):
looks like it works

apps/api/src/services/dining.ts
HwijungK: Wait disregard this comment, doesn't make sense

apps/api/src/schema/dining.ts
```diff
@@ -4,11 +4,24 @@ export const restaurantIdSchema = z
...
-    description: "Filter events by restaurant ID",
-  }),
-});
+export const diningEventsQuerySchema = z
+  .object({
+    restaurantId: restaurantIdSchema.optional().openapi({
+      description: "Filter events by restaurant ID",
+    }),
+    startDate: z.iso.date().optional().openapi({
+      description: "Only return events whose start date is on or after this date",
```
HwijungK: Append `description` with "if provided,"

HwijungK (commented):
tiny nit, i'll just change for u

HwijungK (changes requested):
now that we give `startDate` and `endDate` I think it makes sense to sort the response by startDate

apps/api/src/rest/routes/dining.ts
```diff
@@ -32,7 +32,7 @@ const eventsRoute = createRoute({
   path: "/events",
   request: { query: diningEventsQuerySchema },
   description:
-    "Retrieves all ongoing and upcoming dining events at the current timestamp. Events with null end time are served for 2 weeks after their removal from the underlying data source,",
```
laggycomputer, line 37: I think we should remove references to "ongoing and upcoming" so that we can simplify this description. Details should be left to query parameter documentation.

apps/api/src/schema/dining.ts
```diff
@@ -4,11 +4,24 @@ export const restaurantIdSchema = z
...
-    description: "Filter events by restaurant ID",
-  }),
-});
+export const diningEventsQuerySchema = z
+  .object({
+    restaurantId: restaurantIdSchema.optional().openapi({
+      description: "Filter events by restaurant ID",
+    }),
+    startDate: z.iso.date().optional().openapi({
+      description: "If provided, only return events whose start date is on or after this date",
```
laggycomputer: On the above, I think the default semantics should be specified here.

laggycomputer (changes requested):

laggycomputer (approved):

apps/api/src/graphql/schema/dining.ts
```diff
@@ -133,6 +133,8 @@ type Schedule @cacheControl(maxAge: 3600) {
 
 input DiningEventsQuery {
   restaurantId: RestaurantId
+  startDate: String
+  endDate: String
```
laggycomputer: can we rename these to before and after

laggycomputer (changes requested):

HwijungK (changes requested):
Order responses by `startDate`/`after` as requested previously

laggycomputer (approved):

HwijungK (approved):

PR #406, opened 2026-05-15
fix: ap exam importer should respect existing data

### LINKED ISSUE #164: AP exam importer should clear `ap_exam` table

iAmbrosial:
## Description

Use new ap exam rewards data while removing stale junction rows.

## Related Issue

#164 

## Motivation and Context

Make data good

## How Has This Been Tested?

Ran ap exams importer

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/ap-exams-importer/package.json
+++ b/apps/data-pipeline/ap-exams-importer/package.json
@@ -9,6 +9,7 @@
     "start": "dotenv -e ../../../.env -- tsx src/index.ts"
   },
   "dependencies": {
-    "@packages/db": "workspace:*"
+    "@packages/db": "workspace:*",
+    "drizzle-orm": "0.45.2"
   }
 }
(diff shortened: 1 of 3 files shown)


apps/data-pipeline/ap-exams-importer/package.json
```diff
@@ -9,6 +9,7 @@
     "start": "dotenv -e ../../../.env -- tsx src/index.ts"
   },
   "dependencies": {
-    "@packages/db": "workspace:*"
+    "@packages/db": "workspace:*",
+    "drizzle-orm": "0.45.2"
```
laggycomputer: do not depend on drizzle :(

laggycomputer (changes requested):

apps/data-pipeline/ap-exams-importer/package.json
iAmbrosial: me when I go brain dead and somehow decide that replicating packages/db/utils is the correct choice

laggycomputer (approved):
at long last

PR #413, opened 2026-05-20
feat(library-traffic): Library Traffic Averaging Endpoints

### LINKED ISSUE #243: Library Traffic Data Averaging Using Historical Data

estao1:
## Description

Adds three new REST + GraphQL endpoints under `/v2/rest/libraryTraffic` for accessing historical occupancy data, alongside small consistency improvements to the existing snapshot endpoint.

### New endpoints

| Endpoint | Purpose |
|---|---|
| `GET /history` | Raw, cursor-paginated occupancy rows. Filter by location, date range, or academic term. |
| `GET /history/aggregated` | Time-bucketed averages (`hour` / `day` / `week` / `month`). Scope with `(startDate + endDate)` or `(year + quarter)`. Range cap per granularity to bound response size. |
| `GET /history/pattern` | Recurring-slot pattern averages — e.g. "typical 2pm", "typical Monday". Granularity controls the cycle (24 hour-of-day buckets, 7 day-of-week, 10 week-of-term, 12 month). |

Equivalent GraphQL queries (`libraryTrafficHistory`, `libraryTrafficHistoryAggregated`, `libraryTrafficHistoryPattern`) are exposed with matching shapes and `@cacheControl(maxAge: 1800)`.

### Implementation notes

- **Calendar-term joins.** `year + quarter + period` are convenience filters that resolve via `calendarTerm`. Raw and aggregated overwrite the date range from the term; pattern joins on `timestamp BETWEEN periodStart AND periodEnd` so each row gets its term context.
- **Week-of-term.** The pattern endpoint computes week buckets relative to the term's period start (`floor((ts - periodStart)/604800) + 1`), so week numbers are 1–10 (term-relative) rather than ISO weeks of the year. When `granularity=week` with an explicit quarter filter, rows are separated per-term and include `year` / `quarter` in the response; otherwise data is combined across terms.
- **Composite cursor.** Raw pagination uses the row's UUID as the cursor with a `(timestamp > cursor.ts) OR (timestamp = cursor.ts AND id > cursor.id)` predicate to keep pagination stable across duplicate timestamps (multiple locations scraped at the same second).
- **Production cache.** REST routes register `productionCache({ maxAge: 1800 })`, matching the 30-minute scraper cadence and the GraphQL `@cacheControl` TTLs.
- **Error mapping.** Invalid cursors and missing calendar terms throw `HTTPException(400)` instead of `Error`, so users get proper status codes instead of 500s.
- **Empty results.** Cursor pagination's natural end state is an empty page — the three new history endpoints return `200` with an empty array/items, matching the codebase convention (courses, instructors, enrollment-history). The pre-existing snapshot endpoint keeps its `400-on-empty` behavior.

## Related Issue

Closes #243

## Motivation and Context

The existing snapshot endpoint only exposes current occupancy. Consumers (study-spot recommendations, dashboards, etc.) need historical views — both raw records for export/analysis and aggregated/pattern views for charts.

## How Has This Been Tested?

Manually verified each endpoint via `curl` against a local dev server with backfilled history:

- `GET /libraryTraffic` — snapshot regression
- `GET /libraryTraffic/history` — pagination terminates correctly; composite cursor handles duplicate timestamps
- `GET /libraryTraffic/history/aggregated` with `(startDate + endDate)`, `(year + quarter)`, and missing both (422)
- `GET /libraryTraffic/history/pattern` with `granularity=hour|day|week|month`, with and without `year`/`quarter` filters, verifying week-of-term buckets are 1–10 and labels are clean (`"2pm"`, `"Monday"`, `"Week 5"`)
- GraphQL queries via Yoga playground for the same matrix
- `pnpm check:types` and `pnpm check:biome` clean for `apps/api`

image
image
image

## Types of changes

## Checklist:


apps/api/src/services/library-traffic.ts
```diff
@@ -173,35 +182,67 @@ export class LibraryTrafficService {
...
+      }
+
+      // Use EXISTS instead of JOIN to avoid summer term overlap double-counting rows.
+      // Summer1/Summer10wk/Summer2 instruction periods overlap, so a JOIN would duplicate
+      // rows for those timestamps and bias the averages.
+      const yearClause = input.year ? sql` AND ct.year = ${input.year}` : sql``;
+      const quarterClause = input.quarter ? sql` AND ct.quarter = ${input.quarter}` : sql``;
+      const existsClause = isFinals
+        ? sql`EXISTS (SELECT 1 FROM calendar_term ct WHERE ${libraryTrafficHistory.timestamp} BETWEEN ct.finals_start AND ct.finals_end${yearClause}${quarterClause})`
+        : sql`EXISTS (SELECT 1 FROM calendar_term ct WHERE ${libraryTrafficHistory.timestamp} BETWEEN ct.instruction_start AND ct.instruction_end${yearClause}${quarterClause})`;
```
laggycomputer: We can't use any less raw SQL for this?

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,5 +1,26 @@
...
+    second: "2-digit",
+    hour12: false,
+  }).format(utcDate);
+  const localStr = la.replace(" ", "T");
+  const offsetMin = (Date.parse(`${localStr}Z`) - utcDate.getTime()) / 60000;
+  const sign = offsetMin >= 0 ? "+" : "-";
+  const abs = Math.abs(offsetMin);
+  const offset = `${sign}${String(Math.floor(abs / 60)).padStart(2, "0")}:${String(abs % 60).padStart(2, "0")}`;
+  return `${localStr}${offset}`;
+}
```
laggycomputer: I suspected this is duplicated with study rooms code. Make sure this is new or consolidate.

laggycomputer (changes requested):

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,5 +1,250 @@
...
+  locationName: z.string().optional().openapi({
+    example: "4th Floor - Nordstrom Honors Study Room",
+    description: "Filter results by name of this floor / section",
+  }),
+  year: z.string().optional().openapi({
+    example: "2025",
+    description: "Academic year — use with quarter to scope results to a term",
+  }),
+  quarter: z
+    .enum(["Fall", "Winter", "Spring", "Summer1", "Summer10wk", "Summer2"])
```
HwijungK: We can use `terms` enum from the schema

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,5 +1,250 @@
...
+    example: 100,
+  }),
+});
+
+export const libraryTrafficHistoryRawEntrySchema = z.object({
+  locationId: z.number().int().openapi({ example: 212 }),
+  locationName: z.string().openapi({ example: "4th Floor - Nordstrom Honors Study Room" }),
+  libraryName: z.string().openapi({ example: "Langson Library" }),
+  trafficCount: z.number().int().nonnegative().openapi({ example: 57 }),
+  trafficPercentage: z.coerce.number().openapi({ example: 0.38 }),
```
HwijungK: Do we need to use `coerce` on the output?

apps/api/src/schema/library-traffic.ts
```diff
@@ -27,6 +272,7 @@ export const libraryTrafficEntrySchema = z.object({
     example: "4th Floor - Nordstrom Honors Study Room",
     description: "Name of this floor / section",
   }),
+  locationId: z.number().int().openapi({ example: 212 }),
```
HwijungK: `locationId` is duplicate of `id` and is causing zod to die on `/LibraryTraffic` calls.

apps/api/src/schema/library-traffic.ts
```diff
@@ -18,7 +263,7 @@ export const libraryTrafficQuerySchema = z.object({
 });
 
 export const libraryTrafficEntrySchema = z.object({
-  id: z.number().int().nonnegative().openapi({ example: 245 }),
+  id: z.number().int().nonnegative().openapi({ example: 212 }),
   libraryName: z.string().openapi({
```
HwijungK: Is there a reason we're opposed to making this an enum. If so, we should at least document what strings `libraryName` can be.

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,5 +1,250 @@
...
+      example: "Winter",
+      description: "Academic quarter — use with year to scope results to a term",
+    }),
+  period: z.enum(["instruction", "finals"]).default("instruction").openapi({
+    description: "Which part of the term to filter to (only applies when year + quarter provided)",
+    example: "instruction",
+  }),
+  startDate: z.coerce.date().optional().openapi({
+    example: "2025-01-01T00:00:00Z",
+    description: "Start of the time range (inclusive) — overridden by year/quarter if provided",
```
HwijungK: In my opinion explicit date ranges should take priority over year/quarter sinse it has the potential to be more fine-grained, and somebody who uses the explict dates probably knows what they're doing.

It might be a better idea to disallow a client from querying both ways, or to simply apply both conditions if both are given.

Open to discussion about this, what r your thoughts?

HwijungK (changes requested):
Gonna have lunch will review more soon

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +74,264 @@ export class LibraryTrafficService {
...
+
+    if (input.locationName) {
+      conds.push(eq(libraryTraffic.locationName, input.locationName));
+    }
+
+    if (input.libraryName) {
+      conds.push(eq(libraryTraffic.libraryName, input.libraryName));
+    }
+
+    if (!startDate || !endDate) {
```
HwijungK: This might be better placed as a `refine` clause inside zod

apps/api/src/schema/library-traffic.ts
estao1: I think applying both conditions is fine, as somebody might want more granular time filtering within a specific quarter

apps/api/src/schema/library-traffic.ts
estao1: confirmed new

apps/api/src/graphql/schema/library-traffic.ts
```diff
@@ -52,7 +43,7 @@ input LibraryTrafficHistoryRawQuery {
   libraryName: String
```
HwijungK: Use enums like REST

apps/api/src/services/library-traffic.ts
HwijungK: #300 adds WEEK_MS to stdlib which we could use instead of a hardcoded number but its not merged yet... If it gets merged before reminders to add. if not, its whatever really

HwijungK, line 192: Its my understanding that `avg` will return a Number anyway and doesn't need the `mapWith`

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +74,281 @@ export class LibraryTrafficService {
...
+    const isWeek = input.granularity === "week";
+    // Week bucket is term-relative (1-10), which only makes sense per-quarter; for other
+    // granularities (hour-of-day, day-of-week, month) we always combine all terms
+    const separateByTerm = isWeek && !!input.quarter;
+
+    // Convert stored UTC timestamps to Pacific time before extracting hour/day/month buckets
+    const localTs = sql`(${libraryTrafficHistory.timestamp} AT TIME ZONE 'UTC' AT TIME ZONE 'America/Los_Angeles')`;
+
+    const bucketExpr = {
+      hour: sql<number>`EXTRACT(hour FROM ${localTs})`,
```
HwijungK: Should we use `mapWith` here instead of sql<number>. So that we don't have to cast these to number later. That seems unintuitive as we're forced to casta even when typescript displays that they are numbers.

Not super sure how these are suppose to work, so if you implemented it this way for a reason, I'd love to know.

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +74,281 @@ export class LibraryTrafficService {
...
+        .from(libraryTrafficHistory)
+        .innerJoin(libraryTraffic, eq(libraryTrafficHistory.locationId, libraryTraffic.id))
+        .where(and(...localConds, inTerm))
+        .groupBy(...groupBy)
+        .orderBy(asc(libraryTrafficHistory.locationId), asc(bucketExpr));
+    })();
+
+    return rows.map((row) => ({
+      ...row,
+      bucket: Number(row.bucket),
```
HwijungK: This is the cast to `number` I'm refering to in the early comment on `selectFields`

HwijungK (changes requested):

apps/api/src/graphql/schema/library-traffic.ts
```diff
@@ -13,7 +13,91 @@ input LibraryTrafficQuery {
...
+type LibraryTrafficHistoryRawEntry @cacheControl(maxAge: 1800) {
+  locationId: Int!
+  locationName: String!
+  libraryName: String!
+  trafficCount: Int!
+  trafficPercentage: Float!
+  timestamp: String!
+}
+
+type LibraryTrafficHistoryRawPage {
```
laggycomputer: Why not cached?

apps/api/src/rest/routes/library-traffic.ts
```diff
@@ -44,4 +56,82 @@ libraryTrafficRouter.openapi(libraryTrafficRoute, async (c) => {
...
+
+const libraryTrafficHistoryAggregatedRoute = createRoute({
+  summary: "Retrieve aggregated historical library traffic data",
+  operationId: "libraryTrafficHistoryAggregated",
+  tags: ["Library Traffic"],
+  method: "get",
+  path: "/history/aggregated",
+  request: { query: libraryTrafficHistoryAggregatedQuerySchema },
+  description:
+    "Retrieves time-averaged occupancy metrics grouped by granularity. Scope with either (startDate + endDate) or (year + quarter). Max range: 14 days (hour), 365 days (day), 730 days (week/month).",
```
laggycomputer: I'm not quite sure what this means. Can it be made clearer or written in a way that referring to the time related query parameters makes the explanation clearer?

apps/api/src/rest/routes/library-traffic.ts
```diff
@@ -44,4 +56,82 @@ libraryTrafficRouter.openapi(libraryTrafficRoute, async (c) => {
...
+
+const libraryTrafficHistoryPatternRoute = createRoute({
+  summary: "Retrieve pattern-averaged library traffic data",
+  operationId: "libraryTrafficHistoryPattern",
+  tags: ["Library Traffic"],
+  method: "get",
+  path: "/history/pattern",
+  request: { query: libraryTrafficHistoryPatternQuerySchema },
+  description:
+    "Retrieves typical occupancy averaged across recurring time slots (e.g. all Mondays, all 2pm hours). Granularity controls the cycle: hour → 24 buckets (0-23), day → 7 (1=Mon...7=Sun), week → 10 buckets (week-of-term), month → 12.",
```
laggycomputer: Let's document this "granularity" on the relevant parameter...

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,251 @@
...
+  const offsetMin = (Date.parse(`${localStr}Z`) - utcDate.getTime()) / 60000;
+  const sign = offsetMin >= 0 ? "+" : "-";
+  const abs = Math.abs(offsetMin);
+  const offset = `${sign}${String(Math.floor(abs / 60)).padStart(2, "0")}:${String(abs % 60).padStart(2, "0")}`;
+  return `${localStr}${offset}`;
+}
+
+const libraryNames = ["Langson Library", "Science Library", "Gateway Study Center"] as const;
+
+const MAX_RANGE_DAYS: Record<string, number> = {
```
laggycomputer: `as const`, please

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,251 @@
...
+    description: "Filter results by name of this floor / section",
+  }),
+  year: z.string().optional().openapi({
+    example: "2025",
+    description: "Academic year — use with quarter to scope results to a term",
+  }),
+  quarter: z.enum(terms).optional().openapi({
+    example: "Winter",
+    description: "Academic quarter — use with year to scope results to a term",
+  }),
```
laggycomputer, line 33: Mostly too obvious. What does it mean to "filter by term"? What are we filtering?

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,251 @@
...
+  period: z.enum(["instruction", "finals"]).default("instruction").openapi({
+    description: "Which part of the term to filter to (only applies when year + quarter provided)",
+    example: "instruction",
+  }),
+  startDate: z.coerce.date().optional().openapi({
+    example: "2025-01-01T00:00:00Z",
```
laggycomputer: I don't think we need an example...

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,251 @@
...
+      example: "Langson Library",
+      description: "Filter results by library name",
+    }),
+    locationName: z.string().optional().openapi({
+      example: "4th Floor - Nordstrom Honors Study Room",
+      description: "Filter results by name of this floor / section",
+    }),
+    granularity: z.enum(["hour", "day", "week", "month"]).openapi({
+      description: "Time bucket size for averaging results",
+      example: "hour",
```
laggycomputer: No need for this example

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,251 @@
...
+    }),
+    year: z.string().optional().openapi({
+      example: "2025",
+      description:
+        "Academic year — use with quarter to scope results to a term (alternative to startDate/endDate)",
```
laggycomputer: We're copy-pasting now; outline and reuse

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,251 @@
...
+      "Academic year for this bucket. Populated from term data when week granularity with quarter filter; otherwise echoes the year query param if provided.",
+  }),
+  quarter: z.string().optional().openapi({
+    example: "Fall",
+    description:
+      "Academic quarter for this bucket. Populated from term data when week granularity with quarter filter; otherwise echoes the quarter query param if provided.",
+  }),
+  bucket: z.number().int().openapi({
+    example: 5,
+    description: "Hour (0-23), ISO day of week (1=Mon-7=Sun), week of term (1-10), or month (1-12)",
```
laggycomputer: This is a bit strange. It seems like this parameter takes on different meanings. Why not just have timestamps or multiple optional fields?

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +74,281 @@ export class LibraryTrafficService {
...
+      locationName: libraryTraffic.locationName,
+      libraryName: libraryTraffic.libraryName,
+      year: separateByTerm ? calendarTerm.year : sql<string | null>`null`,
+      quarter: separateByTerm ? calendarTerm.quarter : sql<string | null>`null`,
+      bucket: bucketExpr,
+      avgCount: avg(libraryTrafficHistory.trafficCount).mapWith(Number),
+      avgPercentage: avg(libraryTrafficHistory.trafficPercentage).mapWith(Number),
+    };
+
+    const rows = await (async () => {
```
laggycomputer: Why an IIFE?

laggycomputer (changes requested):

apps/api/src/schema/library-traffic.ts
HwijungK: I'd like to keep this one actually. Helps people see the expected format of input

laggycomputer: Doesn't it say ISO 8601 on there? People can look that up; very rarely will you create that manually...

apps/api/src/services/library-traffic.ts
estao1: ill keep this open for now then

estao1: i believe drizzle types `avg` as `SQL<string | null>` so we'd need it for the `z.number` parse

apps/api/src/schema/library-traffic.ts
```diff
@@ -225,7 +183,8 @@ export const libraryTrafficHistoryPatternEntrySchema = z.object({
   }),
   bucket: z.number().int().openapi({
     example: 5,
-    description: "Hour (0-23), ISO day of week (1=Mon-7=Sun), week of term (1-10), or month (1-12)",
+    description:
+      "Position within the granularity's cycle (its meaning is set by the request's granularity): hour of day (0-23), ISO day of week (1=Mon-7=Sun), week of term (1-10; Fall 0-10), or month (1-12). See `label` for the human-readable form.",
   }),
   label: z.string().openapi({
     example: "2pm",
```
HwijungK, line 163: Nit: Lets match the example of label and buckets

apps/api/src/rest/routes/library-traffic.ts
```diff
@@ -44,4 +56,82 @@ libraryTrafficRouter.openapi(libraryTrafficRoute, async (c) => {
...
+
+const libraryTrafficHistoryPatternRoute = createRoute({
+  summary: "Retrieve pattern-averaged library traffic data",
+  operationId: "libraryTrafficHistoryPattern",
+  tags: ["Library Traffic"],
+  method: "get",
+  path: "/history/pattern",
+  request: { query: libraryTrafficHistoryPatternQuerySchema },
+  description:
+    "Averages occupancy across recurring time slots to reveal typical patterns (e.g. all Mondays, all 2pm hours). The recurring cycle is set by the `granularity` parameter.",
```
laggycomputer: Bit nebulous. Try a notion of "window" (2-3pm) and "stride" (every Monday).

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
 import { z } from "@hono/zod-openapi";
+import { terms } from "@packages/db/schema";
+
+// Format a UTC Date as ISO 8601 with Pacific time offset (e.g. "2026-11-15T14:00:00-08:00").
+// Handles PST/PDT automatically via IANA timezone data.
+function toPacificISO(utcDate: Date): string {
```
laggycomputer: If we already return ISO 8601, I see no need to make sure it's in Pacific. More code to maintain with little benefit.

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
...
+    description:
+      "Academic year of the term to scope to. Combined with quarter, restricts results to readings recorded during that term.",
+  }),
+  quarter: z.enum(terms).optional().openapi({
+    example: "Fall",
+    description:
+      "Academic quarter of the term to scope to. Combined with year, restricts results to readings recorded during that term.",
+  }),
+  period: z.enum(["instruction", "finals"]).default("instruction").openapi({
+    example: "instruction",
```
laggycomputer: No need for example

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
...
+// Query-parameter fields shared by every history endpoint. Each endpoint extends this base
+// with the fields that differ (granularity, pagination, per-granularity range notes).
+const historyFilterBase = z.object({
+  libraryName: z.enum(libraryNames).optional().openapi({
+    example: "Langson Library",
+    description: "Filter results by library name",
+  }),
+  locationName: z.string().optional().openapi({
+    example: "4th Floor - Nordstrom Honors Study Room",
+    description: "Filter results by name of this floor / section",
```
laggycomputer: Keep to the "location" nomenclature

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
...
+  }),
+  year: z.string().optional().openapi({
+    example: "2026",
+    description:
+      "Academic year of the term to scope to. Combined with quarter, restricts results to readings recorded during that term.",
```
laggycomputer: Should help to remind readers of context. What are we scoping? Also no need to mention what quarter does; it's obvious enough from other parts of the API

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
...
+// keeps fine-grained (e.g. hourly) queries from scanning unbounded history.
+const MAX_RANGE_DAYS = {
+  hour: 14,
+  day: 365,
+  week: 730,
+  month: 730,
+} as const;
+
+// Query-parameter fields shared by every history endpoint. Each endpoint extends this base
+// with the fields that differ (granularity, pagination, per-granularity range notes).
```
laggycomputer: Not necessary

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
...
+  take: z.coerce.number().int().positive().max(500).default(100).openapi({
+    description: "Number of rows to return per page (max 500)",
+    example: 100,
+  }),
+});
+
+export const libraryTrafficHistoryRawEntrySchema = z.object({
+  locationId: z.number().int().openapi({ example: 212 }),
+  locationName: z.string().openapi({ example: "4th Floor - Nordstrom Honors Study Room" }),
+  libraryName: z.enum(libraryNames).openapi({ example: "Langson Library" }),
```
laggycomputer, line 64: Similar to enrollment, is there a way to avoid repeating these?

apps/api/src/schema/library-traffic.ts
```diff
@@ -1,8 +1,210 @@
...
+    .openapi({ example: "2026-01-15T14:00:00-08:00" }),
+});
+
+export const libraryTrafficHistoryRawSchema = z.array(libraryTrafficHistoryRawEntrySchema);
+
+export const libraryTrafficHistoryAggregatedQuerySchema = historyFilterBase
+  .extend({
+    granularity: z.enum(granularities).openapi({
+      description: "Size of each time bucket that readings are averaged into",
+    }),
```
laggycomputer, line 82: I like something like `windowSize` more. What do you think?

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +80,270 @@ export class LibraryTrafficService {
...
+        .from(calendarTerm)
+        .where(and(eq(calendarTerm.year, input.year), eq(calendarTerm.quarter, input.quarter)));
+      if (!term) {
+        throw new HTTPException(400, {
+          message: `No calendar term found for ${input.year} ${input.quarter}`,
+        });
+      }
+      const isFinals = input.period === "finals";
+      // Compare Pacific calendar date so readings on the last day of the term are included
+      // in full — a UTC midnight bound would drop the last ~8h of evening data.
```
laggycomputer: Drop this comment

apps/api/src/services/library-traffic.ts
```diff
@@ -1,10 +1,58 @@
...
+    "November",
+    "December",
+  ][bucket - 1];
+}
+
+// A Drizzle date({ mode: "date" }) value is a JS Date; postgres.js can't bind a Date as a
+// parameter inside a raw sql`` comparison (there's no column type to encode it with), so render
+// it to a YYYY-MM-DD string and cast it in SQL instead. UTC slice is exact since the driver
+// parses date-only values at UTC midnight.
+const toDateString = (d: Date): string => d.toISOString().slice(0, 10);
```
laggycomputer: This should live somewhere more general (db or stdlib package)

The comment will also need to be generalized

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +80,270 @@ export class LibraryTrafficService {
...
+    const isWeek = input.granularity === "week";
+    // Week bucket is term-relative (1-10), which only makes sense per-quarter; for other
+    // granularities (hour-of-day, day-of-week, month) we always combine all terms
+    const separateByTerm = isWeek && !!input.quarter;
+
+    // Convert stored UTC timestamps to Pacific time before extracting hour/day/month buckets
+    const localTs = sql`(${libraryTrafficHistory.timestamp} AT TIME ZONE 'UTC' AT TIME ZONE 'America/Los_Angeles')`;
+
+    // mapWith(Number) decodes these numeric SQL results to JS numbers, so callers (and the
+    // groupBy/orderBy/label code below) get real numbers without re-casting each row.
```
laggycomputer: No need for this

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +80,270 @@ export class LibraryTrafficService {
...
+    const bucketExpr = {
+      hour: sql`EXTRACT(hour FROM ${localTs})`.mapWith(Number),
+      day: sql`EXTRACT(isodow FROM ${localTs})`.mapWith(Number),
+      // Week of term: calendar-date difference in Pacific time, 1-indexed.
+      // Using Pacific dates (not raw UTC instants) matches the /week endpoint convention and
+      // avoids mis-attributing late-Sunday-Pacific readings to the following week.
+      // Fall starts on Thursday so subtract 4 days so that Monday of the first full week = day 0,
+      // meaning the Thu–Sun stub is "week 0" (matching /week) and Mon of week 1 = day 7.
```
laggycomputer: Instead of re-explaining this, you can reference the week logic from other endpoints and use that to base your argument

apps/api/src/services/library-traffic.ts
```diff
@@ -32,4 +80,270 @@ export class LibraryTrafficService {
...
+      year: separateByTerm ? calendarTerm.year : sql<string | null>`null`,
+      quarter: separateByTerm ? calendarTerm.quarter : sql<string | null>`null`,
+      bucket: bucketExpr,
+      avgCount: avg(libraryTrafficHistory.trafficCount).mapWith(Number),
+      avgPercentage: avg(libraryTrafficHistory.trafficPercentage).mapWith(Number),
+    };
+
+    // Default (non-week) path: EXISTS instead of JOIN avoids summer term overlap double-counting.
+    // Summer1/Summer10wk/Summer2 instruction periods overlap, so a JOIN would duplicate rows for
+    // those timestamps and bias the averages.
```
laggycomputer: This isn't on only one path. Move this comment or whatever else is necessary to make it correct.

laggycomputer (changes requested):

apps/api/src/services/library-traffic.ts
```diff
@@ -158,17 +161,21 @@ export class LibraryTrafficService {
...
-        - CASE WHEN ${calendarTerm.quarter} = 'Fall' THEN INTERVAL '4 days' ELSE INTERVAL '0 days' END
-      )) / 604800)::int + 1`,
-      month: sql<number>`EXTRACT(month FROM ${localTs})`,
+      hour: sql`EXTRACT(hour FROM ${localTs})`.mapWith(Number),
+      day: sql`EXTRACT(isodow FROM ${localTs})`.mapWith(Number),
+      // Week of term: calendar-date difference in Pacific time, 1-indexed.
+      // Using Pacific dates (not raw UTC instants) matches the /week endpoint convention and
+      // avoids mis-attributing late-Sunday-Pacific readings to the following week.
+      // Fall starts on Thursday so subtract 4 days so that Monday of the first full week = day 0,
+      // meaning the Thu–Sun stub is "week 0" (matching /week) and Mon of week 1 = day 7.
```
HwijungK: goofed comment ('Mon of week 1 = day 7')

apps/api/src/services/library-traffic.ts
```diff
@@ -158,17 +161,21 @@ export class LibraryTrafficService {
...
+      week: sql`floor(
+        (${localTs}::date - ${periodStart}
```
HwijungK, line 165: This causes queries with period = isFinal to mis-lable finals weeks as "week 0" and "week 1"

apps/api/src/schema/library-traffic.ts
estao1: i think granularity is the more standard term in similar time-series APIs like Grafana

HwijungK (changes requested):
Pattern Endpoint fails

Issue #416, opened 2026-05-22
Allow providing department and course number instead of "Course ID"


laggycomputer:
We should add (department, course number) to catalogue side stuff and ~~course ID to WebSoc sided stuff~~ done in #419.

leonwang1016:
Hello, just to confirm, has this task already been completed?

HwijungK:
This task was already done in #419. Issue was left opened by mistake

HwijungK:
@leonwang1016 You could take a look at https://github.com/icssc/anteater-api/issues/428  if you would like

PR #419, opened 2026-05-23
feat(websoc): support courseId query param

KevinWu098:
## Description

Adds optional `courseId` to the WebSoc query endpoint (REST + GraphQL), filtering on the existing generated `websoc_course.course_id` column (`REPLACE(dept_code, ' ', '') || course_number`). 

Also returns `courseId` on each course in the response so clients can use the same identifier as catalogue, programs, and `/websoc/syllabi`.

## Related Issue

#416

## Motivation and Context

Clients already identify courses by `courseId` (e.g. `I&CSCIH32`) elsewhere in the API. The main WebSoc query previously required `department` + `courseNumber`; this adds a single-key filter aligned with the syllabi endpoint.

For AntAlmanac, this allows better querying for collections of courses (which are more idiomatically identified by a courseId than a set of tuples which have to be serialized/deserialized between AA and AAPI).

## How Has This Been Tested?

Not manually tested. Logic mirrors `getSyllabi` (`eq(websocCourse.courseId, …)`); no DB migration.

## Types of changes

## Checklist:


apps/api/src/graphql/schema/websoc.ts
```diff
@@ -50,6 +50,7 @@ type WebsocSection @cacheControl(maxAge: 300) {
 
 type WebsocCourse @cacheControl(maxAge: 300) {
     sections: [WebsocSection!]!
+    courseId: String!
```
HwijungK: Do we really need to give both courseId and dept + num as a field in the response. Maybe its good practice, but it feels redundant to me

apps/api/src/schema/websoc.ts
HwijungK, line 93: Nit: It makes good UX sense to put courseNumber right after department. Probably order ... dept, courseNum, courseId, courseTitle, ...

HwijungK (changes requested):

apps/api/src/schema/websoc.ts
```diff
@@ -86,6 +86,10 @@ export const websocQuerySchema = z.object({
     description: "Only include courses offered by the specified department code",
     example: "I&C SCI",
   }),
+  courseId: z.string().optional().openapi({
+    description: "The course ID to query",
```
HwijungK: Docs change to reflect that user should use courseId OR Dept + name, as per Dante

HwijungK (changes requested):

apps/api/src/graphql/schema/websoc.ts
KevinWu098: I think... it's nice because it will let us do less string concatenation in app land

HwijungK (approved):
I wont fight on the docs... this is an ok from me.

apps/api/src/schema/websoc.ts
```diff
@@ -86,13 +86,17 @@ export const websocQuerySchema = z.object({
     description: "Only include courses offered by the specified department code",
     example: "I&C SCI",
   }),
+  courseNumber: courseNumberSchema.optional().openapi({
+    description: "Only include courses with the specified course number",
+  }),
+  courseId: z.string().optional().openapi({
+    description: "The course ID to query.",
```
laggycomputer: reframe in terms of "only include..." or add "mutually exclusive with..."

laggycomputer (changes requested):

laggycomputer (changes requested):
not in favor of the reordering of fields because this disagrees with other API surface areas (e.g. LARC, other parts of websoc)

otherwise cool

KevinWu098:
image

laggycomputer (approved):
run it

Issue #409, opened 2026-05-18
Dining: restrict historical schedule data


laggycomputer:
Introduced by #392.

We should find a way to limit how much data comes back when `includeHistorical` is specified.
Maybe a window (no more than 2 weeks old from today), or perhaps query parameters.

PR #424, opened 2026-05-28
Dining time bound historical schedules

HwijungK:
## Description

In addition to old `includeHistorical` parameter that determines whether the endpoint will return all schedules or only those that have not come to pass, We now offer `before` and `after` parameters. 

`before` queries all schedules which began before a certain date
`after` queries all schedules which ended after a certain date
`before` and `after` can be used in conjunction to query windows of time

`includeHistorical` is only considered when neither `before` nor `after` is given and will return schedules that have not passed.

All queries return schedules with `null` end dates, which is the default schedule.

## Related Issue

#409 

## Motivation and Context

Previously we had if we wanted to see historical data, we had to retrieve every saved schedule. This allows users to limit the amount of data they recieve around the time they are interested in.

## How Has This Been Tested?

Runs locally on REST and GraphQL endpoints.

## Types of changes

## Checklist:


laggycomputer (changes requested):
this is based on #300

HwijungK:
> based on #300

Ye Im fixing that rn TT, gonna just copy it over....

PR #425, opened 2026-05-28
feat (dining): restrict past schedules by before and after dates

### LINKED ISSUE #409: Dining: restrict historical schedule data

HwijungK:
## Description

In addition to old `includeHistorical` parameter that determines whether the endpoint will return all schedules or only those that have not come to pass, We now offer `before` and `after` parameters. 

`before` queries all schedules which began before a certain date
`after` queries all schedules which ended after a certain date
`before` and `after` can be used in conjunction to query windows of time

`includeHistorical` is only considered when neither `before` nor `after` is given and will return schedules that have not passed.

All queries return schedules with `null` end dates, which is the default schedule.

## Related Issue

#409 

## Motivation and Context

Previously we had if we wanted to see historical data, we had to retrieve every saved schedule. This allows users to limit the amount of data they recieve around the time they are interested in.

## How Has This Been Tested?

Runs locally on REST and GraphQL endpoints.

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/services/dining.ts
+++ b/apps/api/src/services/dining.ts
@@ -7,6 +7,7 @@ import {
   inArray,
   isNull,
   lt,
+  lte,
   max,
   min,
   or,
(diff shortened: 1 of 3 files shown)


apps/api/src/services/dining.ts
```diff
@@ -297,11 +298,20 @@ export class DiningService {
...
+    }
+    if (query.after) {
+      dateRangeConds.push(gte(diningSchedule.endDate, query.after));
+    }
+    if (!query.after && !query.before && !query.includeHistorical) {
+      dateRangeConds.push(gte(diningSchedule.endDate, sql`CURRENT_DATE`));
+    }
+
+    if (dateRangeConds.length) {
+      conds.push(or(isNull(diningSchedule.endDate), and(...dateRangeConds))!);
```
laggycomputer: Our lints should forbid `!`. Did they not trigger?

apps/api/src/services/dining.ts
```diff
@@ -297,11 +298,20 @@ export class DiningService {
       conds.push(eq(diningSchedule.restaurantId, query.restaurantId));
     }
 
-    if (!query.includeHistorical) {
-      // standard and current/upcoming special schedules by default
-      conds.push(
-        or(isNull(diningSchedule.endDate), gte(diningSchedule.endDate, sql`CURRENT_DATE`))!,
-      );
+    const dateRangeConds: SQL[] = [];
```
laggycomputer, line 301: You may safely type this as `(SQL | undefined)[]`, and it might help with the non-null assertion below...

laggycomputer (changes requested):

laggycomputer (approved):

PR #432, opened 2026-06-02
fix (dining): Make start time of dining events nullable

HwijungK:
## Description

Scraper was dying because the pk on `dining_event` was a multi-column pk on `restaurant_id`, `title`, and `start`, but the amnesty week event had a null start date for some reason...

If we remove the `start`, `restaurant_id` and `title` would not be sufficient enough to serve as a primary key because of events that repeat over years. Ended up adding a uuid for pk instead and dropping the original primary key.

We added a unique constraint on `restaurant_id`, `start`, and `end`. This is under the assumption that if a event is found with the same start and end dates in the same restaurant, it is a rename of a previously listed event and not a new one that runs simultaneously.

We don't know if there are other events with null start and end times other then "Amnesty Weeks ". For such cases, we cannot upsert with a unique condition on `restaurant_id`, `start`, and `end` (with NULLDISTINCT other events that have null start and end times will override each other & with NULL NOT DISTINCT we would add copies of the same event on every scrape). 

We added a partial unique index on `restaurantId` and `title` instead and use that as the target for cases of null start times.

## How Has This Been Tested?
Scrape works now

## Checklist:


packages/db/src/schema.ts
```diff
@@ -1048,14 +1049,10 @@ export const diningEvent = pgTable(
...
-    return {
-      pk: primaryKey({
-        name: "dining_event_pk",
-        columns: [table.title, table.restaurantId, table.start],
-      }),
-    };
-  },
+  (table) => [
+    // We assume that restaurants cannot have event with the same start and end times, and such cases are simply a rename of the previous event.
+    unique().on(table.restaurantId, table.start, table.end).nullsNotDistinct(),
```
laggycomputer: amnesty from this year will conflict with next year
what's the play

laggycomputer (changes requested):

packages/db/src/schema.ts
HwijungK: your right it should be distinct.

Also, it doesn't really make sense the amnesty year should have a null start and end, which implies that there may be times when other events are also like this. Those shouldn't override each other.

laggycomputer (approved):
make sure you test this all the way to endpoint layer but i'm satisfied

PR #436, opened 2026-06-07
fix(degreeworks-scraper): deduplicate by major {code->name}

HwijungK:
When we get the list of potential programs (schoolCode, degreeCode, collegeCode, majorCode, specCode), DW often returns duplicates.
when scraping for major programs, we save each scraped audit as <Major_Name>;<Spec_Code> (i.e "Major in Computer Science;201B", "Major in Computer Science", etc)
This works for current catalogs. However, from #344 which saves historical catalog years, we discover that historically, there is no name differentiation between majors of the same subject with by different degrees.
For example, in the year 20132014, both MS in Math and BS in Math were labeled as "Major In Mathematics". This caused the scraper to only scrape the first program, and mistakenly skip the others as they had the same `majorSpecId` (current catalog saves MS-540 as "MS in Mathematics" which is why wasn't noticed before).

We changed `majorSpecId` to save <Degree_Code>-<Major_Code>[;<Spec_Code>]. (ie BS-201;201B)

Scrape of catalog year 20132014 found 16 new programs (86 -> 102)
Scrape of catalog year 20252026 remained the same(266)

## Types of changes

## Checklist:

### DIFF
--- a/apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
+++ b/apps/data-pipeline/degreeworks-scraper/src/components/Scraper.ts
@@ -21,8 +21,8 @@ import {
 
 const JWT_HEADER_PREFIX_LENGTH = 7;
 
-// 'majorName;specCode'. If no specialization is taken, simply 'majorName'
-type MajorSpecId = string | `${string};${string}`;
+// 'degreeCode-majorCode;specCode'. If no specialization is taken, simply 'degreeCode-majorCode'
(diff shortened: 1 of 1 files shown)


HwijungK:
Update ChatGPT claims that there number of programs went from 196 to 231 from 2013 to 2025. 196 is very far off from 102, so more investigation should be done on this

laggycomputer (approved):
data anomalies aside this is Just Correct so do it

Issue #418, opened 2026-05-22
Docs (Websoc): Provide meaning of course codes


HwijungK:
Websoc has many 1 Letter restriction codes that mean stuff. We should provide in the `excludeRestrictionCode` and `restriction` fields documentation for what each letter means. This information can be found on websoc.

PR #437, opened 2026-06-09
docs: Provide meaning of course codes

limlindsayj:
## Description
Added openapi descriptions to excludeRestrictionCode and restriction. 

## Related Issue
#418 

## Motivation and Context
Currently there is no description or context for these in the schema. It is helpful to have a list of what these restriction codes mean in the description, or you would have to look it up yourself in websoc. 

## Types of changes

## Checklist:


laggycomputer (changes requested):
- Not sure why Wrangler is being addled with again (I'm sure something strange happened). Please revert.
- We're pasting in a somewhat critical string in two different places. Try not to repeat it: use a constant string, shared Zod schema, OpenAPI type reference, or something else.
- I don't exactly want to be responsible for the ground truth. What are your thoughts on just linking the WebSoc pages which lists these meanings?

Issue #444, opened 2026-07-13
Consistency in sample program names


laggycomputer:
These are properly "catalogue program" and "catalog program variations". 
The SQL and Drizzle names should reflect this.

leonwang1016:
Hello!
I created a pull request for this issue. Could you review it? Thanks!

PR #446, opened 2026-07-17
fix: rename sample_program_variation to catalogue_program_variation

leonwang1016:
…(#444)

## Description

Renamed the `sample_program_variation` table to `catalogue_program_variation` and the `sample_program` column to `catalogue_program` to be consistent with the existing `catalogue_program` table naming convention.

## Related Issue

[

](https://github.com/icssc/anteater-api/issues/444)

## Motivation and Context

The table and column names were inconsistent with the rest of the schema. This change makes the naming uniform.

## How Has This Been Tested?

Verified that all references to the old names were updated in the Drizzle schema, services, and a new migration file was added.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code involves a change to the database schema.


HwijungK (changes requested):
There's some changes to `apps/key-manager/cloudflare-env.d.t.s` that i don't think you intended to do. Heads up when you call `pnpm i` the hash changes in that file but that's safe to just ignore and keep out of the pr. Not sure what removing `GlobalProps` was.

Additionally, when you create a change to the schema, instead of manually writing a migration file to `/migration` you should run the `npm run db:generate -- --name="<NAME>`. This additionally creates a snapshot in `snapshot/` and adds an entry to `_journal` (read https://orm.drizzle.team/docs/drizzle-kit-generate and https://orm.drizzle.team/docs/drizzle-kit-migrate). when you want to run the migration call `npm run db:migrate`

packages/db/src/schema.ts
```diff
@@ -233,13 +233,13 @@ export const catalogProgram = pgTable("catalogue_program", {
...
 
-export const sampleProgramVariation = pgTable("sample_program_variation", {
+export const catalogueProgramVariation = pgTable("catalogue_program_variation", {
   id: uuid("id").primaryKey().defaultRandom(),
   programId: varchar("program_id")
     .notNull()
     .references(() => catalogProgram.id, { onDelete: "cascade" }),
   label: varchar("label"),
-  sampleProgram: jsonb("sample_program").$type<SampleProgramEntry[]>().notNull(),
+  catalogueProgram: jsonb("catalogue_program").$type<SampleProgramEntry[]>().notNull(),
```
HwijungK: rename to type to `CatalogProgramEntry`

apps/key-manager/cloudflare-env.d.ts
HwijungK, line 1: You forgot to remove this file.

packages/db/src/schema.ts
```diff
@@ -233,13 +233,13 @@ export const catalogProgram = pgTable("catalogue_program", {
   programName: varchar("program_name").notNull(),
 });
 
-export const sampleProgramVariation = pgTable("sample_program_variation", {
+export const catalogueProgramVariation = pgTable("catalogue_program_variation", {
```
HwijungK: lets call this `catalogProgramVariation` to match `catalogProgram`

HwijungK (changes requested):
Hitting endpoint locally returns a 5XX error, investigate.

When we make changes you should always test the local endpoint, and run the relevant scraper locally to make sure that they both work (find and test the catalog program scrapper too).

Also look at the relevant graphql/schema and see if there's any matching changes you should do.

leonwang1016:
Just checking, after my most recent commit, is my pull request ready? Or am I missing something. Thanks!

HwijungK (changes requested):
make sure you uncomment out the keyverifier and accessController lines u ignored for testing.

Also, I usually make several passes through code reviews and find bugs that I missed the time before. I didn't have time to review if it works (it should probably) so I'll do another check today or tmr, agian. The PR will probably be ready after that

packages/db/src/schema.ts
```diff
@@ -220,26 +220,26 @@ export type StandingYearType = (typeof StandingYear)[number];
...
-export type SampleProgramEntry = {
+export type CatalogProgramEntry = {
   year: StandingYearType;
   fall: CourseEntry[];
   winter: CourseEntry[];
   spring: CourseEntry[];
 };
 
-// Sample Programs Tables
+// Catalogue Program Tables Tables
```
HwijungK: typo

apps/api/src/graphql/schema/catalogue.ts
```diff
@@ -23,23 +23,23 @@ type YearPlan @cacheControl(maxAge: 86400) {
     spring: [CourseEntry!]!
 }
 
-type SampleProgramVariation @cacheControl(maxAge: 86400) {
+type CatalogProgramVariation @cacheControl(maxAge: 86400) {
```
HwijungK: Changed my mind. I feel graphql/schema/ should still use sample programs. This is b/c I feel like the user should see 'sample programs' which makes more sense from its name then 'catalog program'. Lets use 'catalog' for the db schema and the scraper/index, while keepining the api side using sampleProgram

HwijungK (changes requested):
1. If you haven't already, do `git merge main` to sync this branch to main. We did some major dependency updates (i.e TypeScript 5.9 -> 6.0), which is probably y your having trouble with TS and the werid `cloudflare-env.d.ts` changes. After you merge, run `pnpm i` again to update dependency on your end. This should update the `hash` in cloudflare-env.d.ts but they can be saftely ignored (do a `git revert` on the file)

2. If you create migration files in a pr, and then need to ammend the migration (i.e you renamed to `cataloguePrograms` and created migration files, but later decide you should actually have used `catalogPrograms`), you should delete the previous migration .sql file, snapshot file, and the corresponding entry from `_journal.json`, then create a new migration which would keep all ur pr changes to 1 migration.

3. Currently endpoint is failing, consider what SQL command your trying to run on the database.

HwijungK (approved):
Lgtm!

PR #449, opened 2026-07-19
Fix/439 prereq split

### LINKED ISSUE #439: Prerequisite checking splits too aggressively
it's RAW!

https://github.com/icssc/anteater-api/blob/d4ffa873ef9133ae64435ce78675a1b4448d0116/apps/data-pipeline/course-scraper/src/index.ts#L239

leonwang1016:
## Description

Changed the prerequisite list splitting regex from `/AND/` to `/\bAND\b/` (word boundary) to correctly split on the standalone word "AND" regardless of surrounding whitespace, while avoiding false splits on "AND" appearing inside words (e.g. "DANCE", "STANDARD").

The word boundary `\b` approach correctly handles all three spacing variations:
- `WRITING AND LOWER` (spaces on both sides)
- `WRITINGAND LOWER` (space only on right)
- `WRITING ANDLOWER` (space only on left)

## Related Issue

  https://github.com/icssc/anteater-api/issues/439 

## Motivation and Context

The original `/AND/` regex was too aggressive and would split on "AND" appearing anywhere in a string. The initial fix of `/ AND /` was too strict and missed cases where "AND" appeared without spaces on one or both sides (e.g. "WRITINGAND", "ANDLOWER"), which are common in the raw HTML from the UCI catalogue.

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/apps/data-pipeline/course-scraper/package.json
+++ b/apps/data-pipeline/course-scraper/package.json
@@ -14,7 +14,6 @@
     "cheerio": "1.2.0",
     "cross-fetch": "4.1.0",
     "domhandler": "6.0.1",
-    "entities": "^8.0.0",
     "json-diff": "1.0.6",
     "readline-sync": "1.4.10",
     "sort-keys": "6.0.1",
--- a/apps/data-pipeline/course-scraper/package.json
+++ b/apps/data-pipeline/course-scraper/package.json
@@ -14,6 +14,7 @@
     "cheerio": "1.2.0",
     "cross-fetch": "4.1.0",
     "domhandler": "6.0.1",
+    "entities": "^8.0.0",
     "json-diff": "1.0.6",
     "readline-sync": "1.4.10",
     "sort-keys": "6.0.1",
--- a/apps/data-pipeline/course-scraper/src/index.ts
+++ b/apps/data-pipeline/course-scraper/src/index.ts
@@ -236,7 +236,7 @@ function buildORLeaf(prereqTree: PrerequisiteTree, prereq: string) {
 
 function buildPrereqTree(prereqList: string): PrerequisiteTree {
   const prereqTree: PrerequisiteTree = { AND: [], NOT: [] };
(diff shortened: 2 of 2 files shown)


HwijungK (changes requested):
The scraper crashes when i try to run it. Always smoke test your changes, and remove any debugging code from the pr.

leonwang1016:
Hello
I removed the debugging code that previously caused the scraper to crash before it could even be tested.
I have tested the scraper itself and believe the HTML implementation works in **scrapePrerequisitePage**.

cheerio.text() removes the HTML structure around 
image
, causing prerequisite tokens to be concatenated (e.g. WRITINGAND, ANDLOWER). Extracting the HTML and stripping tags preserves the intended whitespace while keeping the visible text intact.

00f0da7b-0b4e-4c59-b5bb-cbc97b9a8a9f

This allows / AND /  in **buildPrereqTree** to work.

HwijungK (changes requested):
Run another test where you compare the results of previous /AND/ splitting and current / AND / splitting (after replacing html tags from `.html()` output). We expect to see differences for prereq text that had `STANDING`, `DANCE`, etc. Are there any other changes and are they correct?

leonwang1016:
Hello
Found some differences between /AND/ and / AND /
image

It seems like / AND / is properly splitting now unlike /AND/ which split the word **STANDING**.

leonwang1016:
Ran a temporary comparison logger (\bAND\b vs / AND /) across the full scraped dataset. Zero differences found, meaning there shouldn't be any unintended changes beside the word AND being embedded inside an actual valid word.

leonwang1016:
image

image

These screenshots show the comparison between the original pipeline (text extraction and /AND/ split) vs the new pipeline (html tag stripped and / AND /).

HwijungK (changes requested):
your leaving extra spaces in.

make sure to remove debug code.

HwijungK (approved):
lgtm

Issue #428, opened 2026-05-29
Filter Courses by Terms Offered


Ascent817:
## Expected Behavior

The `/v2/rest/search` endpoint should accept an optional, comma-separated `terms` query parameter that restricts course results to those historically offered in at least one of the given quarters the past year, like this:

```
GET /v2/rest/search?query=algorithms&resultType=course&terms=Fall,Spring
```

- Accepts the existing `term` enum values: `Fall`, `Winter`, `Spring`, `Summer1`, `Summer10wk`, `Summer2`.
- OR across multiple terms (like how `ge` currently works)

## Current Behavior

The search endpoint supports `department`, `courseLevel`, `minUnits`/`maxUnits`, and `ge`, but has no way to filter by the quarters a course is offered. Each course result already includes a `terms` array, but with no server-side filter a consumer can only filter the current page of results client-side, which breaks the `count` and pagination (`skip`/`take`).

## Context

In Planner, students plan around when a course is typically offered ("what can I take in Spring?"). A server-side quarter filter would make this much easier to do.

laggycomputer:
It has been decided that PPC's unique use case is better served from within their own code.

PR #451, opened 2026-07-24
feat: add terms filter to search endpoint (#428)

leonwang1016:
## Description

 Added an optional `terms` query parameter to the `/v2/rest/search` endpoint. It accepts a comma-separated string of terms (e.g. "Fall,Spring") and filters course search results to only return courses that have been offered in at least one of the specified terms.

## Related Issue

 https://github.com/icssc/anteater-api/issues/428 

## Motivation and Context

 Users had no way to filter search results by which quarter a course is offered in. This change allows users to narrow down search results to courses offered in specific quarters, which is useful for schedule planning.

## How Has This Been Tested?

 Tested locally using the API reference page at localhost:8787/reference. Searched for "algorithms" with resultType=course and terms=Fall. Verified that all returned courses had at least one Fall term in their terms array. Also verified that an invalid term value returns a 422 validation error.

image
image

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/src/graphql/schema/search.ts
+++ b/apps/api/src/graphql/schema/search.ts
@@ -26,6 +26,7 @@ input SearchQuery {
     minUnits: Float
     maxUnits: Float
     ge: [String!]
+    terms: [Term!]
 }
 
 extend type Query {
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -9,6 +9,7 @@ import {
   globalRateLimiter,
   headerInjector,
   ipBasedRateLimiter,
+  keyVerifier,
   openapiMeta,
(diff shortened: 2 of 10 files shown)


laggycomputer (changes requested):
- Ground truth for the list of terms exists elsewhere already and should be used.
- Consideration needs to be given for use of indexes in this query. Please either show that indexes are invoked, show that indexes are not necessary and the compute cost is acceptable without them, or add appropriate indexes.

laggycomputer (changes requested):
- There are presumably development-oriented changes committed. These should not be in the patch.
- GraphQL has been neglected. Please make sure GraphQL can access the same (or reasonably equivalent) functionality.

PR #454, opened 2026-07-25
fix(key-manager): provide `client_id` when authing

HwijungK:
adding the `client` property with `token_endpoint_auth_method: 'none'` sends client_id as part of the body instead of as an encoded header because nextauth ✨magic✨

I can find 0 good docs on how this works but these are some bread crumbs
https://github.com/nextauthjs/next-auth/issues/5122
https://github.com/panva/openid-client/blob/main/docs/functions/None.md

This will make key-manager a public client instead of confidential. I think this is intentional because our auth expects client_id to be in the body.

## How Has This Been Tested?

Ran local Auth and Key-manager. I think it works

## Types of changes

## Checklist:


apps/key-manager/src/auth.ts
```diff
@@ -19,7 +19,6 @@ declare module "next-auth" {
 export const { handlers, signIn, signOut, auth } = NextAuth(async (_req) => {
   const cf = getCloudflareContext();
   const db = database(cf.env.USERS_DB.connectionString);
-
```
laggycomputer, line 22: revert

laggycomputer (changes requested):

laggycomputer (changes requested):
conventional commits

laggycomputer (approved):

PR #467, opened 2026-08-13
refactor(key-manager): type guards

laggycomputer:
waow

### DIFF
--- a/apps/key-manager/src/app/actions/keys.ts
+++ b/apps/key-manager/src/app/actions/keys.ts
@@ -40,7 +40,7 @@ export async function keyFromFormToStorage(
   return newDataAsStorage;
 }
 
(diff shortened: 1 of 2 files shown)


HwijungK (approved):
storage looks fine to me (i hate it).

Some questions about keys.ts sinse i didn't review that for the previous prs. Should we standardize what we call a `key` and a `keyId` and can we rename `createKeyInner` to be something more informative?

laggycomputer:
> Should we standardize what we call a `key` and a `keyId`

nomenclature is kinda weird because the string used to give you special access to the API is both a key and an ID referencing data in kv storage

the barrier i tried to enforce is that it's a key id whenever the additional data is involved and a key otherwise

> can we rename `createKeyInner` to be something more informative?

ok

PR #474, opened 2026-09-05
Fix(DegreeWorks): max shared in nonexclusive qualifier

HwijungK:
THe bugs never stopr


apps/api/src/schema/programs.ts
```diff
@@ -117,6 +117,9 @@ export const nonExclusiveQualifierSchema = z
               "The code of a specific program this qualifier applies to. These are ids for majors, minors, and specializations (i.e. `BS-201`, `120`, `BS-201A`); numerical codes representing colleges for college requirements (i.e `55` for School of Biological Sciences); and misc strings for other blocks (i.e `LIBL` for Liberal Learning). If no code is specified, this qualifier applies to all programs of the specified `programType`",
             examples: ["BS-201", "120", "BS-201A", "55", "LIBL"],
           }),
+        maxShared: z.string().optional().openapi({
+          description: "If given, the maximum number of courses that can be shared.",
```
laggycomputer, line 121: be precise "maximum number of courses in this block which may apply to another applicable block simultaneously"

laggycomputer, line 121: why not int

laggycomputer (changes requested):

Issue #438, opened 2026-06-10
Prerequisite data drops non-course clauses (class standing, school/major requirements)


dsnsgithub:
## Context

I noticed COMPSCI 184A's WebSoc restriction code is just `A: Prerequisite required`, while the actual requirements (upper division standing, ICS/CSE majors) are prerequisites. Since AAPI drops those clauses, AntAlmanac Scheduler/Planner do not have a reliable way to show them, and students discover the restrictions only when enrollment fails.

<img width="2455" height="805" alt="Image" src="https://github.com/user-attachments/assets/874f1f81-f01d-4a80-9b92-b28ae21caafc" />

<img height="105" alt="Image" src="https://github.com/user-attachments/assets/06294cce-36d0-4050-a95f-844ce959f944" />

For COMPSCI 184A, the registrar lists ([source](https://www.reg.uci.edu/cob/prrqcgi?term=202692&dept=COMPSCI&action=view_by_term#184A)):

<img width="2185" height="265" alt="Image" src="https://github.com/user-attachments/assets/0e2caf03-c7e5-48c4-bcbf-765c4e61ebeb" />

## Current Behavior / Steps to Reproduce
 [GET /v2/rest/courses/COMPSCI184A](https://anteaterapi.com/v2/rest/courses/COMPSCI184A) currently returns:

```json
"prerequisiteTree": {"AND": [{"OR": [
    {"prereqType": "course", "coreq": false, "courseId": "I&C SCI 6N", "minGrade": "D-"},
    {"prereqType": "course", "coreq": false, "courseId": "MATH 3A", "minGrade": "D-"}
]}]},
"prerequisiteText": "I&C SCI 6N or MATH 3A"
```

---

## Possible Solution (Claude)

Add a third variant to the `prereq` schema alongside `course` and `exam`:

```json
{
    "prereqType": "requirement",
    "requirement": "UPPER DIVISION STANDING ONLY",
}
```

- We may or may not need to add `"examName": "UPPER DIVISION STANDING ONLY"` for backwards compatibility 🤮
- partially caused by: https://github.com/icssc/anteater-api/issues/439

## Your Environment
- Anteater API

PR #479, opened 2026-09-12
Feat: 438 prerequisite requirement clauses

leonwang1016:
## Description

Fixed prerequisite data loss bug and added classification of non-course requirement text from WebSoc.

1. Replaced reqWithGradeMatch and courseCoreqMatch in the parsePrerequisite function with a new separate function called parseAnnotatedCourseOrExam that can handle courses with both ( coreq ) and (min grade = X ) at the same time (ex: DRAMA 10). Previously those courses were discarded. 
2. Modified parseAntirequisite to also handle non course/exam requirement exclusions.

3. The buildORLeaf function now uses recursion to handle 

( AC ENG 20A OR ( PLACEMENT EXAM OR AUTHORIZATION (see SOC comments for authorization policy/instructions) ) )

by splitting PLACEMENT EXAM from AUTHORIZATION.

4. Parsed the requirements into "status", "qualification", or "completion" and extracted them.
- FRESHMAN, SOPHOMORE, JUNIOR, SENIOR, UPPER DIVISON, LOWER DIVISION, GRADUATE, CAMPUSWIDE HONORS, SCHOOL, and MAJORS go into the "status" category
- SAT/ACT scores, PLACEMENT EXAM, and AUTHORIZATION go into the "qualification" category."
- LOWER DIVISION WRITING and ENTRY LEVEL WRITING go into the "completion" category.
- Anything else that doesn't fit into these three categories is stored as plain text. There was one example of this: "Bio 199 Packet must be submitted to Bio Sci Student Affairs prior to enrollment."

5. Some courses such as PSYCH173A have incomplete prereqs on WebSoc so added loggers to flag them. isBalancedPrereqText does a simple check for this.
image

6. buildPrereqTree pipeline: 
- The entire requirement string is first split by splitOnAnd, which looks for AND delimeters
- For each piece, if it starts with "(", it goes into the splitOnOr branch, which looks for OR delimeters.

## Related Issue

 https://github.com/icssc/anteater-api/issues/438 

## Motivation and Context

1. Previously, courses such as DRAMA 10 ( coreq ) ( min grade = D- ) were being silently discarded instead of being parsed. 
2. Non course requirement clauses were discarded instead of being parsed.
## How Has This Been Tested?

Below is the json file containing each course, its raw text, and the extracted.
[requirement-clauses.json](https://github.com/user-attachments/files/32481611/requirement-clauses.json)

[1789174024145.log](https://github.com/user-attachments/files/32135307/1789174024145.log)
The above logger text file shows the dropped courses. There are only "recommended", truncated, and "no repeat allowed" courses on there which is intentional.

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [X ] My code involves a change to the database schema.


laggycomputer (changes requested):
Thanks for the PR. A lot of things going on here.

- It is not valid in Conventional Commits to combine verbs as you have done in this PR title. Choose one (the correct answer is that this is a feature).
- You should rebase this PR onto main. I expect this should not create conflicts because this is already based upon recent major changes.
- There should be no changes to any Wrangler-generated TypeScript files.
- There should be OpenAPI examples for added fields, if they are not trivially self-documenting like enums and bools, and documentation where it can be helpful beyond existing context.
- I appreciate the logs and output from a successful run, but an important piece of data for auditing is the input which was provided. One good format would be a CSV.
- Without that data, assuming your conception of the requirements is correct, more can be done to add enums, tagged unions, etc. to create a stronger contract about what requirements will look like.
- On the above, consider the following patterns you can utilize:
  - Freshman/sophomore/junior/senior are standing levels.
  - School/major/CHC are affiliations.
  - AP, IB, and placement exams are exams.
  - This probably continues.

I'll look at the code once we come to an understanding on these points. Please feel free to ask questions to flesh this out.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +295,162 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
+  }
+
+  if (current.trim()) parts.push(current.trim());
+  //logger.info(`splitOnOr input: ${JSON.stringify(prereqList)}`);
+  //logger.info(`splitOnOr output: ${JSON.stringify(parts)}`);
+  return parts;
 }
 
 function buildPrereqTree(prereqList: string): PrerequisiteTree {
+  if (BOILERPLATE_STRINGS.some((s) => prereqList.includes(s))) {
```
HwijungK: The boilerplate_string is an invalid bit that is scraped because it happens to share the same "table tbody tr" shape that cheerio scrapes for. if you try to resolve the bug like this you're going to have an empty 'course' with said name in the prerequistes map.

Doesn't affect the end result but its still a bug

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +295,162 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
+  let current = "";
+
+  for (let i = 0; i < prereqList.length; i++) {
+    const char = prereqList[i];
+
+    if (char === "(") depth++;
+    else if (char === ")") depth--;
+
+    //word boundary check so words like "JUNIOR" and "SENIOR" don't get split on accident
+    const precededByLetter = i > 0 && /[A-Za-z]/.test(prereqList[i - 1]);
```
HwijungK: seems cleaner to check simply for if \s is before unless there's a case where it doesn't work like that

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +174,124 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
+    if (gradeMatch) {
+      if (gradeMatch[1] === "grade") {
+        minGrade = gradeMatch[2];
+      } else {
+        isExam = true;
+        minGrade = gradeMatch[2];
+      }
+      continue;
+    }
+    return undefined;
```
HwijungK, line 206: is there a reason to return undefined here? looks like a silent failure

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +174,124 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
+  }
+
+  if (isExam) {
+    return { prereqType: "exam", examName: base, ...(minGrade ? { minGrade } : {}) };
+  }
+  return coreq
```
HwijungK, line 212: this can be simplified

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +174,124 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
     return prereq.startsWith("AP")
       ? { prereqType: "exam", examName: prereq }
       : { prereqType: "course", coreq: false, courseId: prereq };
   }
+
+  if (/^PLACEMENT EXAM$/i.test(prereq)) {
+    return { prereqType: "exam", examName: prereq };
+  }
+
+  const satActMatch = prereq.match(/^((?:SAT|ACT)\s+.+?)\s*(?:>=|<=|=)\s*(\d+)$/i);
```
HwijungK: is `<=` really nessessary?

HwijungK (changes requested):

laggycomputer:
I'm seeing a lot of updates here. Make sure that you resolve comments and request review as necessary to advance the PR further. We are available to help, but you must be proactive.

apps/data-pipeline/course-scraper/src/index.ts
leonwang1016: an unrecognized annotation would reach this step and then be dropped, can i add a logger.warn here so it doesn't get silently discarded?

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,73 @@ export const coursesByCursorQuerySchema = z.object({
   }),
 });
 
+const standingPrerequisiteSchema = z.union([
```
laggycomputer: Use a discriminated union.

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,73 @@ export const coursesByCursorQuerySchema = z.object({
...
+        "SOPHOMORE",
+        "JUNIOR",
+        "SENIOR",
+        "LOWER DIVISION",
+        "UPPER DIVISION",
+        "GRADUATE",
+        "NEW TRANSFERS",
+      ]),
+    })
+    .strict(),
```
laggycomputer: There is never a need for this in this context, because we would never write extra fields. Remove everywhere.

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,73 @@ export const coursesByCursorQuerySchema = z.object({
   }),
 });
 
+const standingPrerequisiteSchema = z.union([
+  z
```
laggycomputer: All of these can use OpenAPI descriptions.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +177,143 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
   }
 }
 
+/*
+replaces old reqWithGradeMatch and courseCoreqMatch with this one function that can handle courses with both such as DRAMA 10 ( coreq ) ( minGrade )
```
laggycomputer: This is appropriate for a PR comment or other explanatory context outside the codebase, but is not helpful in the codebase.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +177,143 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
+    }
+    logger.warn(
+      `Unrecognized annotation [${JSON.stringify(ann)}] in ${JSON.stringify(prereq)}; dropping entire match.`,
+    );
+    return undefined;
+  }
+  if (isExam) {
+    return { prereqType: "exam", examName: base, ...(minGrade ? { minGrade } : {}) };
+  }
+  return { prereqType: "course", coreq, courseId: base, ...(minGrade ? { minGrade } : {}) };
```
laggycomputer: Is there actually a downside to preventing `{ minGrade: undefined }`?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +177,143 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
-  }
-  const courseCoreqMatch = prereq.match(/^([^()]+)\s+\( coreq \)$/);
-  if (courseCoreqMatch) {
-    return { prereqType: "course", coreq: true, courseId: courseCoreqMatch[1].trim() };
+  if (/\(\s*recommended\s*\)/i.test(prereq)) {
+    //logger.info(`IGNORING RECOMMENDED PREREQUISITE: ${JSON.stringify(prereq)}`);
+    return undefined;
   }
+
+  const annotated = parseAnnotatedCourseOrExam(prereq);
```
laggycomputer: This variable name is no longer appropriate.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +177,143 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
+
+  const schoolMatch = text.match(/^SCHOOL OF (.+?)\s+ONLY$/i);
+
+  if (schoolMatch) {
+    return {
+      prereqType: "affiliation",
+      school: text
+        .replace(/^SCHOOL OF\s+/i, "")
+        .replace(/\s+ONLY$/i, "")
+        .trim(),
```
laggycomputer: Why are capture groups not used here?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +177,143 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
+    };
+  }
+
+  const majorMatch = text.match(/^(.+?)\s+MAJORS?\s+ONLY$/i);
+
+  if (majorMatch) {
+    return {
+      prereqType: "affiliation",
+      major: text.replace(/\s+MAJORS?\s+ONLY$/i, "").trim(),
+    };
```
laggycomputer: Why are capture groups not used here?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
   if (antiCourseMatch) {
     return { prereqType: "course", coreq: false, courseId: antiCourseMatch[1].trim() };
   }
+
+  const withoutNo = prereq.replace(/^NO\s+/, "").trim();
+
+  //call parseAnnotatedCourseOrExam to handle ( coreq ) and ( min grade = C ) annotations
```
laggycomputer: You may specify what you mean by an "annotation" somewhere, but otherwise this comment is a statement of the plainly obvious.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
+  const annotated = parseAnnotatedCourseOrExam(withoutNo);
+  if (annotated) {
+    return annotated;
+  }
+
+  //already handled by parseRepeatability
```
laggycomputer: Add a space after `//`.

laggycomputer: It would be helpful to formally define this behavior in writing. Think about what this statement means with respect to handling the potential conflict between WebSoc and catalogue data.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
+  if (/^NO REPEATS ALLOWED\b/.test(prereq)) {
+    return undefined;
+  }
+
+  //ex: NO PSYCHOLOGY MAJORS ONLY
+  const extracted = extractStandingOrAffiliation(withoutNo);
+  if (extracted) {
+    /*logger.info(
+      `NEGATED REQUIREMENT CAPTURED [${extracted.category}: ${extracted.value}]: ${JSON.stringify(prereq)}`,
+    );*/
```
laggycomputer: Do you intend to log this?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
   } else {
     const req = parsePrerequisite(prereq);
+    //logger.info(`AND LEAF PARSED: ${JSON.stringify(req)}`);
     if (req) {
       prereqTree.AND?.push(req);
-    }
+    } /*else {
+      logger.warn(`DROPPED AND-LEAF: ${JSON.stringify(prereq)}`);
+    }*/
   }
```
laggycomputer: Consider removing these. They are obtrusive at the moment.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
 }
-
+//uses recursion to handle cases like ( AC ENG 20A OR ( PLACEMENT EXAM OR AUTHORIZATION (see SOC comments for authorization policy/instructions) ) )
```
laggycomputer: Space after `//`.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
 function buildORLeaf(prereqTree: PrerequisiteTree, prereq: string) {
+  //logger.info(`PREREQ INPUT: ${JSON.stringify(prereq)}`);
+
+  if (prereq.startsWith("(") && prereq.endsWith(")")) {
+    const nestedTree: PrerequisiteTree = { OR: [] };
+    const orReqs = splitOnOr(prereq.slice(1, -1).trim());
+    for (const orReq of orReqs) {
+      buildORLeaf(nestedTree, orReq.trim());
+    }
+    if (nestedTree.OR?.length) {
```
laggycomputer: This shouldn't be fallible, given the definition above, and incurs an extra check which can be avoided. You can probably achieve this by allowing an automatic upcast on the recursive invocation, instead of performing erasure on declaration.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
+      prereqTree.OR?.push(nestedTree.OR.length === 1 ? nestedTree.OR[0] : nestedTree);
+    }
+    return;
+  }
   const req: Prerequisite | undefined = prereq.startsWith("NO")
     ? parseAntirequisite(prereq)
     : parsePrerequisite(prereq);
+
+  //logger.info(`PARSED RESULT: ${JSON.stringify(req)}`);
+
```
laggycomputer: Do you intend to keep this?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,48 +329,155 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
   if (req) {
     prereqTree.OR?.push(req);
+  } /*else {
+    logger.warn(`Undefined prerequisite parsed in buildORLeaf: ${JSON.stringify(prereq)}`);
+  }*/
```
laggycomputer: Do you intend to keep this?

laggycomputer (changes requested):
There is enough here to work on for now. I or June will continue reviewing the parsing code later.

apps/data-pipeline/course-scraper/src/index.ts
leonwang1016: i think the simple return statement would also be fine, JSON.stringify omits undefined object properties anyways

apps/data-pipeline/course-scraper/src/index.ts
leonwang1016: Right. There's another case of this in buildPrereqTree.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -573,10 +828,10 @@ async function scrapeCoursesInDepartment(meta: {
   } else {
     console.log(`Difference between database and scraped prerequisite data for ${deptCode}:`);
     console.log(prereqDiff);
-    if (!readlineSync.keyInYNStrict("Is this ok")) {
+    /*if (!readlineSync.keyInYNStrict("Is this ok")) {
```
laggycomputer: Comments like this need to be removed.

packages/db/src/schema/courses.ts
```diff
@@ -33,7 +34,55 @@ export type ExamPrerequisite = {
...
+        type: "major";
+        major: string;
+      }
+    | {
+        type: "school";
+        school: string;
+      }
+    | {
+        type: "honors";
+        honors: true;
```
laggycomputer: This is strange. A runtime value which does not vary does not need to exist at runtime. See https://en.wikipedia.org/wiki/Unit_type.

packages/db/src/schema/courses.ts
```diff
@@ -33,7 +34,55 @@ export type ExamPrerequisite = {
...
+        writingRequirement: WritingRequirement;
+      };
+};
+
+export type AffiliationPrerequisite = {
+  prereqType: "affiliation";
+  affiliation:
+    | {
+        type: "major";
+        major: string;
```
laggycomputer: There's no need to change the field name here, in my opinion. It makes certain dynamic dispatch tricks harder for little reason. Repeat everywhere.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -424,19 +661,38 @@ function parseRepeatability(repeatText: string): {
...
+    return tree.AND.flatMap(toEntry);
   }
   if (tree.OR) {
-    return tree.OR.flatMap((x) => (isPrereq(x) ? prereqToString(x) : prereqTreeToList(x)));
+    return tree.OR.flatMap(toEntry);
   }
   return [];
 }
 
+//check for cliff-hanging source text like for PSYCH 173A
```
laggycomputer: This needs to be explained further. Comments need `// `.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -424,19 +661,38 @@ function parseRepeatability(repeatText: string): {
 
 const isPrereq = (x: Prerequisite | PrerequisiteTree): x is Prerequisite => "prereqType" in x;
 
-const prereqToString = (prereq: Prerequisite) =>
-  prereq.prereqType === "course" ? prereq.courseId.replaceAll(/ /g, "") : prereq.examName;
+//requirement leaves never reach here
+const prereqToString = (
+  prereq: Exclude<Prerequisite, StandingPrerequisite | AffiliationPrerequisite>,
+) => (prereq.prereqType === "course" ? prereq.courseId.replaceAll(/ /g, "") : prereq.examName);
```
laggycomputer: What is a "requirement leaf" which can never reach here? Be more precise.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -424,19 +661,38 @@ function parseRepeatability(repeatText: string): {
 
 const isPrereq = (x: Prerequisite | PrerequisiteTree): x is Prerequisite => "prereqType" in x;
 
-const prereqToString = (prereq: Prerequisite) =>
-  prereq.prereqType === "course" ? prereq.courseId.replaceAll(/ /g, "") : prereq.examName;
+//requirement leaves never reach here
+const prereqToString = (
```
laggycomputer: Arrow functions at top-level are less readable and, I argue, have more unclear semantics. Use a named function.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -424,19 +661,38 @@ function parseRepeatability(repeatText: string): {
...
+  prereq: Exclude<Prerequisite, StandingPrerequisite | AffiliationPrerequisite>,
+) => (prereq.prereqType === "course" ? prereq.courseId.replaceAll(/ /g, "") : prereq.examName);
 
 function prereqTreeToList(tree: PrerequisiteTree): string[] {
+  const toEntry = (x: Prerequisite | PrerequisiteTree): string[] => {
+    if (!isPrereq(x)) return prereqTreeToList(x);
+    return x.prereqType === "standing" || x.prereqType === "affiliation" ? [] : [prereqToString(x)];
+  };
```
laggycomputer: Inline arrows are wasteful if they don't enclose any name from the outer scope. (They can also be wasteful otherwise.) Outline this.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -292,9 +513,25 @@ async function scrapePrerequisitePage(deptCode: string, url: string) {
       if (courseId.match(/\* ([&A-Z\d ]+) since/)) {
         courseId = courseId.split("*")[0].trim();
       }
+      if (BOILERPLATE_STRINGS.some((s) => prereqList.includes(s))) return;
+      if (!isBalancedPrereqText(prereqList)) {
+        logger.warn(
+          `Truncated prereq source for ${courseId}: unbalanced parentheses, likely cut off ` +
+            `by the registrar's page. Skipping. ` +
+            `Raw text: ${JSON.stringify(prereqList)}`,
```
laggycomputer: That's not the raw string, is it? The raw string should not be JSON in my understanding

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -292,9 +513,25 @@ async function scrapePrerequisitePage(deptCode: string, url: string) {
       if (courseId.match(/\* ([&A-Z\d ]+) since/)) {
         courseId = courseId.split("*")[0].trim();
       }
+      if (BOILERPLATE_STRINGS.some((s) => prereqList.includes(s))) return;
```
laggycomputer: Why is this newly necessary? Does this signal a new parsing bug?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -264,24 +482,27 @@ function buildPrereqTree(prereqList: string): PrerequisiteTree {
...
   const prereqs = new Map<string, PrerequisiteTree>();
+  const skippedCourseIds: string[] = [];
   $("table tbody tr").each(function () {
     const entry = $(this).find("td");
     if ($(entry).length === 3) {
       let courseId = $(entry[prereqFieldLabels.Course]).text().replace(/\s+/g, " ").trim();
       const courseTitle = $(entry[prereqFieldLabels.Title]).text().replace(/\s+/g, " ").trim();
-      const prereqList = $(entry[prereqFieldLabels.Prerequisite])
+      const prereqCell = $(entry[prereqFieldLabels.Prerequisite]);
+      const prereqList = prereqCell
```
laggycomputer: If you won't use the first binding, don't create it.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -292,9 +513,25 @@ async function scrapePrerequisitePage(deptCode: string, url: string) {
...
+        );
+        skippedCourseIds.push(courseId);
+        return;
+      }
       prereqs.set(courseId, buildPrereqTree(prereqList));
     }
   });
+  if (skippedCourseIds.length) {
+    logger.warn(
+      `${deptCode}: skipped ${skippedCourseIds.length} course(s) with truncated prerequisite ` +
```
laggycomputer: Why is this section newly needed?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -264,24 +482,27 @@ function buildPrereqTree(prereqList: string): PrerequisiteTree {
       prereqTree.NOT = undefined;
     }
   }
-  return {
+  const resolved: PrerequisiteTree = {
     ...(prereqTree.AND?.length && { AND: prereqTree.AND }),
     ...(prereqTree.OR?.length && { OR: prereqTree.OR }),
     ...(prereqTree.NOT?.length && { NOT: prereqTree.NOT }),
   };
+  return resolved;
```
laggycomputer: If you wanted to cast the output, that could be done inline, though I don't think you need to because it would automatically coerce to the explicitly declared signature. Is there a reason to outline this?

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -209,6 +343,27 @@ function parseAntirequisite(prereq: string): Prerequisite | undefined {
...
     return { prereqType: "course", coreq: false, courseId: antiCourseMatch[1].trim() };
   }
+
+  const withoutNo = prereq.replace(/^NO\s+/, "").trim();
+
+  const annotated = parseAnnotatedCourseOrExam(withoutNo);
+  if (annotated) {
+    return annotated;
+  }
+
```
laggycomputer: Space here.

apps/data-pipeline/course-scraper/src/index.ts
```diff
@@ -169,31 +177,157 @@ async function fetchWithDelay(url: string, delayMs = 1000) {
...
+    };
+  }
+
+  if (/^AUTHORIZATION\b/i.test(prereq)) {
+    return undefined;
+  }
+
+  const extracted = extractStandingOrAffiliation(prereq);
+
+  return extracted;
```
laggycomputer: Why is this outlined?

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,90 @@ export const coursesByCursorQuerySchema = z.object({
...
+          "JUNIOR",
+          "SENIOR",
+          "LOWER DIVISION",
+          "UPPER DIVISION",
+          "GRADUATE",
+          "NEW TRANSFERS",
+        ])
+        .openapi({
+          description:
+            "The class standing required to take this course. Everything except Graduate and New Transfers are unit based.",
```
laggycomputer: There is no need to define the bases of these values, except by linking to a relevant source. This nears a statement of the obvious.

When revising these, please refer to the previewed REST docs to see if it makes sense to you.

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,90 @@ export const coursesByCursorQuerySchema = z.object({
...
+          example: "JUNIOR",
```
laggycomputer: Enums usually don't need examples, and this one doesn't.

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,90 @@ export const coursesByCursorQuerySchema = z.object({
...
+    }),
+    z
+      .object({
+        type: z.literal("writingRequirement"),
+        writingRequirement: z.enum(["LOWER DIVISION WRITING", "ENTRY LEVEL WRITING"]),
+      })
+      .openapi({
+        description:
+          "The writing requirement that must've been completed prior to taking this course.",
+        example: "LOWER DIVISION WRITING",
```
laggycomputer: This provides little value beyond the listing of enum values.

apps/api/src/schema/courses.ts
```diff
@@ -140,23 +140,90 @@ export const coursesByCursorQuerySchema = z.object({
...
+        description: "Students must be in this school to enroll in this course.",
+        example: "I&C SCI",
+      }),
+    }),
+    z.object({
+      type: z.literal("honors"),
+      honors: z.literal(true).openapi({
+        description:
+          "Always true, only appears when students must be in the Campuswide Honors Program.",
+      }),
```
laggycomputer: The documentation you write here should make it obvious that it should not be here. A zero-sized type does not need to be represented at all because it already appears alongisde a discriminant.

laggycomputer (changes requested):
I have more comments, but these are fixes that need to be addressed (including from previous rounds) because they distract me from the actual logic.

**Please resolve comments if you are done with them.**

